PR #72, opened 2014-12-16
padded blocks

industrial-sloth:
This PR implements two main changes:
* it introduces a `PaddedBlocks` class, which acts like the existing `SimpleBlocks` but for an additional padding region of redundant voxels around the edge of each block
* when a tuple of int is passed into a Simple (or Padded) BlockingStrategy to explicitly specify the way in which to divide up an image, this tuple will now by default be interpreted as a block size in pixels, rather than a number of splits to be applied to that dimension of the image. The splitting behavior can be maintained by passing `units="splits"` as an additional argument to `toBlocks` or any of the family of methods that end up calling `toBlocks`.

I'll attach in comments a demo script showing usage of the new padded blocks, to be used with thunder-submit.


python/thunder/rdds/series.py
```diff
@@ -636,6 +636,111 @@ def query(self, inds, var='inds', order='F', onebased=True):
...
+                return iter([(label, val)])
+
+        writer = getParallelWriterForPath(outputdirname)(outputdirname, overwrite=overwrite)
+
+        binseriesrdd = self.rdd.mapPartitions(partitionToBinarySeries)
+
+        binseriesrdd.foreach(writer.writerFcn)
+        # note that we use len(self.index) here for nvalues! we don't really have a great alternative
+        # to this at the moment, short of calling first(). 
+        writeSeriesConfig(outputdirname, len(self.dims), len(self.index), dims=self.dims.count,
```
freeman-lab: We might not want to write `dims` in this config file, for two reasons. One, it potentially triggers a count during the writing stage, which is annoying if that's not something the user cares about storing. Second, on some user-derived `Series` objects I can imagine the dims not being particularly meaningful or useful (e.g. for non-contiguous keys after filtering). In contrast, when doing a conversion from images (by calling `saveAsBinarySeries` on an `Images` object) the dims are always meaningful and useful, so it makes sense to store them. At the very least, I'd say we should only save if it's been computed already. And some point we could add a more generic `nrecords` metadata tag, which would always be stored, and `dims` only if available.

python/thunder/rdds/series.py
```diff
@@ -636,6 +636,111 @@ def query(self, inds, var='inds', order='F', onebased=True):
...
+                          keytype='int16', valuetype=self.dtype, overwrite=overwrite)
```
freeman-lab: One practical consequence of using `int16` for the keys is that the resulting files can basically only be read by us. You cannot really load one of these things in, e.g., numpy or matlab, because those can't handle non-constant numerical type specs (as far as I know). We might not need to fix that... but if we wanted to, we could update the keys to the type of the values before writing. Thoughts?

python/thunder/rdds/series.py
```diff
@@ -636,6 +636,111 @@ def query(self, inds, var='inds', order='F', onebased=True):
...
+        returnType = blockingStrategy.getBlocksClass()
+        blockedRdd = self.rdd.map(blockingStrategy.blockingFunction)
+        # since our blocks are likely pretty big, try setting 1 partition per block
+        groupedRdd = blockedRdd.groupByKey(numPartitions=blockingStrategy.nblocks)
+        # <key>, <val> at this point is:
+        # <block number>, <[(series key, series val), (series key, series val), ...]>
+        simpleBlocksRdd = groupedRdd.map(blockingStrategy.combiningFunction)
+        return returnType(simpleBlocksRdd, dims=self.dims, nimages=len(self.index), dtype=self.dtype)
+
+    def saveAsBinarySeries(self, outputdirname, overwrite=False):
```
freeman-lab: Any thoughts on renaming this to `saveAsBinary`? The disadvantage is that it won't match the corresponding method on `Images`, which is `saveAsBinarySeries`. The advantage is that it avoids the conceptual redundancy inherent in `Series.saveAsBinarySeries` =)

python/thunder/rdds/series.py
```diff
@@ -636,6 +636,111 @@ def query(self, inds, var='inds', order='F', onebased=True):
...
+        """Writes out Series-formatted data.
+
+        This method (Series.saveAsBinarySeries) writes out binary series files using the current partitioning
+        of this Series object. (That is, if mySeries.rdd.getNumPartitions() == 5, then 5 files will be written
+        out, one per partition.) The records will not be resorted; the file names for each partition will be
+        taken from the key of the first Series record in that partition. If the Series object is already
+        sorted and no records have been removed by filtering, then the resulting output should be equivalent
+        to what one would get from calling myImages.saveAsBinarySeries().
```
freeman-lab: Maybe add a note that if users are using this to save a series after it's been converted from images with no other steps in between, they are better off just calling `tsc.convertImagesToSeries`? From some informal testing I think the latter is faster (and that makes sense given the underlying operations, namely the skipping of the `flatMap` into individuated records).

freeman-lab:
@industrial-sloth This is terrific work! I just spent some time doing both local testing and testing on the cluster. Performance was terrific; for a `Series` that took 34 s to load (~100GB), it was 54 s to write back. And everything seemed intact when reloading. One thing that caught me off guard was a doubling of partitions upon re-reading, but that's because the numerical type was `float32` after writing, but `int16` in the original, so thus twice the size / number of partitions, which is the correct behavior. Also spent time testing the various conversions (to blocks, to padded blocks, from series to blocks), and all seemed to work nicely.

freeman-lab:
Beyond the few comments I left (which would be great to address), LGTM!

python/thunder/rdds/series.py
industrial-sloth: Ok, two thoughts, one fairly concrete and the other more theoretical:

* `dims` is currently referenced in the config in two places, once as `dims.count` to get the dimension tuple itself, and the other as `len(dims)` to get the number of keys. Number of keys we do have to have here, because otherwise the data is unreadable, but it isn't stored anywhere else in the Series metadata. So our alternative, if `dims` isn't set on the series object, would be to get `nkeys` from a `series.first()` call. Now this is still better than recalculating the full `dims`, since `dims` requires a full map/reduce cycle through the data, but in all likelihood it still won't be instantaneous. (Another plus is that by calling `first`, we'd get to set `nrecords` directly, rather than by assuming that the `.index` attribute is correct.) So I think I'm in favor of this overall, but read on...

* The Series binary format and logic is growing increasingly hairy, for what was originally I think intended and expected to be a really simple nearly-flat-file representation of the data. We're now proposing to make one of the metadata fields (`dims`) optional. This is not in and of itself a big deal, but it does suggest, at least to me, that at some point we may wish to stop and revisit the entire issue of a persisted representation of Series objects, and see if we can't come up with some more principled, generic, and reusable format, rather than continuing to make ad-hoc changes. (For some reason I keep wanting to say "protocol buffers! avro! thrift!" here, but I'm by no means convinced that's a good idea either. :) )

Takeaway: sure, I'll make `dims` optional, just so long as we're aware that we still have to make a `first()` call in this case.

python/thunder/rdds/series.py
industrial-sloth: Thoughts. Yes. So despite my rant above about the Series binary format becoming unexpectedly complex, I don't see forcing keys and values to be the same data type to be a useful simplification (that's why I changed it in the first place :) ). 

*Writing* one of these files from matlab is still simple-ish - you concatenate your keys and your values, write it out as a gigantic array, and, oh, oops, write out a JSON file specifying that both keys and values are float64. (I'm not aware of an easy way to read or write JSON from matlab? From python of course it's trivial.)

So the issue comes in if someone wanted to use matlab to open up a Series data file that we produced. In return for an extra 144 bytes / record (assuming 3d data saved as float64) and the aesthetic ugliness of representing integer indices as floats, we gain the ability to slurp up this data as a single `read` call or whatever, after which we still have to split the array into keys and values according to an `nkeys` parameter read from JSON. In contrast, as it currently stands, one would have to read this data either through a bunch of `fread` calls specifying the appropriate data type, or possibly still by slurping it as a byte array or something and doing some clever casting afterwards. In numpy this can be read pretty easily with the `struct` module, which is how we're reading and writing it ourselves.

I guess I really don't see this as a useful or cost-effective simplification. Series data is I think well past the point where we can hope to represent it as a single flat array - it's got keys, values, metadata, all kinds of stuff. If we want people to be able to easily read and write it from matlab or numpy, possibly we should be providing our own file i/o routines to be used. And if we're doing that, we should potentially also revisit the format as a whole, as I suggested above, and see if we can't find some more principled and general way to represent all of this.

python/thunder/rdds/series.py
industrial-sloth: I've been thinking of `Binary Series` as *almost* a separate data type from a regular `Series`. In the file output code, series partitions first get mapped from a format where a single record is a 1d array to one where records are these big concatenated binary blobs of keys and values for the whole partition. This second blob representation (plus metadata) I've been thinking of as a "binary series". This is as opposed to, say, writing out a series as a sequence file, where the sequence file abstraction lets you think of the data as being serialized with their format relatively unchanged.

So from that point of view where "binary series" is a thing, it's not redundant. :)

python/thunder/rdds/series.py
industrial-sloth: Sounds good, will add a "see also" or similar. Agreed, calling `convertImagesToSeries` should be at least somewhat faster.

industrial-sloth:
Awesome, thanks Jeremy, glad you like it! I wonder how much of the difference between the 34 s read and the 54 s write is due solely to the data expansion from int16 to float32? Might be interesting (ok, not *very* interesting, but still) to repeat that read/write cycle starting with the floats and see whether it's more comparable at that point? I can't offhand think of anything else that would account for such a pronounced difference in reading and writing the same data...

industrial-sloth:
Hi @freeman-lab - have pushed the changes to only write out the `dims` attribute if already available and added a comment suggesting `convertImagesToSeries` on `series.saveAsBinarySeries`. I still like the `saveAsBinarySeries` name over `saveAsBinary` despite the potential redundancy, and am not thrilled about enforcing the same data type across keys and values, but would be happy to make both those further changes if you'd like. Let me know if you'd want to see either change, but otherwise I think we may be done with this PR!

python/thunder/rdds/series.py
freeman-lab: All good points, great to think it through, and I agree with the choice.

python/thunder/rdds/series.py
freeman-lab: Agree on getting `nkeys` from a first call, and also using the first call to get the `nrecords`. So that's a great change all around.

Also agree in general about standardizing the Series file format some more (see a comment on this below). But with regards to `dims`, I'm pretty sure currently, whether or not `dims` is in the `conf.json` file, it's not used to set that piece of metadata. We may want to change that, and use it to set the metadata if it's present. But the key point is that `dims` is not strictly required to load Series data (in contrast to, say, `nvalues`). So I don't see this as making the metadata itself optional, it's about storing by default only what's required to load it in, and storing other things optionally to provide useful information (and maybe avoid subsequent operations post-loading).

We can continue this discussion offline =)

python/thunder/rdds/series.py
freeman-lab: A couple quick points: 
- there's a simple JSON package in matlab, so writing Series from Matlab or Numpy is easy (assuming one knows the spec). I've had some users doing exactly this. Ultimately, this is my primary reason to prefer the flat binary; when I last investigated more Hadoop friendly formats it was hard to find one that could be trivially written from within those packages.
- I really like the idea of something like a Data API in the future, basically a set of bundled routines to export data from alternate packages into Thunder's Series format, and load them in. Good to know the Python version of such a thing could read something with heterogenous type. I'm not sure the Matlab version could. But let's leave this as is for now and keep thinking about it.

python/thunder/rdds/series.py
industrial-sloth: *checking...*

Wow, you seem to be entirely correct - `dims` isn't used on loading even if it **is** present in the json! Gah! How about we just take this out then? If it isn't used on loading (and I can't offhand think of anything else that would look at `conf.json`) then shouldn't it not be there at all? 

I don't think this changes any of the existing logic in the PR as it currently stands (after the changes I pushed this morning), since if we **do** have `series._dims` set we can still avoid a `first()` call by using it to set `nkeys`, but I can remove the `dims` keyword parameter from `seriesloader.writeSeriesConfig` and thereby hope to head off any further similar confusion in the future.

freeman-lab:
Sorted out some stuff offline with @industrial-sloth regarding when to write `dims` offline with @industrial-sloth . Now LGTM!

industrial-sloth:
Merged from new_imageblocks into master.

PR #74, opened 2014-12-21
Refactored registration

freeman-lab:
Significant refactoring of registration methods based on offline discussions with @broxtronix and @poolio and Noah (@d-v-b might be interested as well).

The key change is to separate the application and estimation of registration parameters through a new `RegisterModel` class. Performing registration will return an instantiation of this class, and it has a `transform` method for applying the transformations to an `Images` object. This makes it simple to compute parameters on one data set and apply them to another, which was previously difficult.

Other features for improved modularity and extensibility
- Support simple saving / loading of registration models to / from JSON
- Separate particular registration implementations into modules within an `imgprocessing.regmethods` package
- Provide simpler constructor interface through a `Register` class that is separate from the base class `RegisterMethod`
- A new `run` method on registration methods combines the `fit` and `transform` into a single step, which can be more efficient, and matches the old behavior, but does not expose the transformation parameters directly

Example usage
- Compute parameters on one data set and apply to another
```
from thunder import Register

images1 = tsc.loadExample('fish-images')
images2 = images1.medianFilter(2)
ref = images2.first()[1]

model = Register('crosscorr').fit(images2, ref)
alginedimages = model.transform(images1)
```
- Perform registration in one step
```
from thunder import Register

images1 = tsc.loadExample('fish-images')
ref = images1.first()[1]

alignedimages = Register('crosscorr').run(images1, ref)
```
- Save registration model to JSON
```
from thunder import Register

images1 = tsc.loadExample('fish-images')
ref = images1.first()[1]

model = Register('crosscorr').fit(images1, ref)
model.save('/tmp/file.json')

model = Register.load('/tmp/file.json')
```


python/thunder/imgprocessing/transformation.py
```diff
@@ -0,0 +1,58 @@
...
+        Apply an n-dimensional displacement by shifting an image or volume.
+
+        Parameters
+        ----------
+        im : ndarray
+            The image or volume to shift
+        """
+        from scipy.ndimage.interpolation import shift
+
+        return shift(im, map(lambda x: -x, self.delta), mode='nearest')
```
poolio: How would you feel about storing the deltas as numpy arrays instead of tuples? This could clean up a lot of these map calls.

python/thunder/imgprocessing/transformation.py
```diff
@@ -0,0 +1,58 @@
...
+class Transformation(object):
+    """ Base class for transformations """
+
+    def apply(self, im):
+        raise NotImplementedError
+
+
+class Displacement(Transformation):
+
+    def __init__(self, delta=None):
```
poolio: document the dimensions/meaning of delta here

python/thunder/imgprocessing/transformation.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return shift(im, map(lambda x: -x, self.delta), mode='nearest')
+
+    def __repr__(self):
+        return "Displacement(delta=%s)" % str(self.delta)
+
+
+class PlanarDisplacement(Transformation):
+
+    def __init__(self, delta=None):
```
poolio: How is delta structured? Would be nice to document it here so you don't have to cross-reference the registration method to figure it out.

python/thunder/imgprocessing/transformation.py
```diff
@@ -0,0 +1,58 @@
+""" Transformations produced by registration methods """
+
+
+class Transformation(object):
```
poolio: What do you think about adding something like a `getParams()` and `setParams()` method that would be used to serialize the Transformation objects?

python/thunder/imgprocessing/register.py
```diff
@@ -1,252 +1,316 @@
...
+        f = open(file, 'r')
+        input = json.loads(f.read())
+
+        # import the appropriate transformation class
+        regmethod = str(input['regmethod'])
+        classname = str(input['transclass'])
+        transclass = getattr(importlib.import_module('thunder.imgprocessing.transformation'), classname)
+
+        # instantiate the transformations and construct the model
+        transformations = {int(k): transclass(**v) for k, v in input['transformations'].iteritems()}
```
poolio: This would require that each `Transform` object can take its `__dict__` as an argument, right? You could overload the default `JSONEncoder` in `Transform` to use `getParams()` and use something like `setParams()` to recreate the object.

python/thunder/imgprocessing/register.py
```diff
@@ -1,252 +1,316 @@
...
-    def filter(self, im):
-        """
-        Apply filtering, and if not set, return image unchanged
-        """
-
-        if hasattr(self, '_filter'):
-            return self._filter(im)
-        else:
-            return im
+    def getTransform(self, im, ref):
```
poolio: Moving `ref` to an argument of `__init__` may make more sense as some registration methods may not use references.

poolio:
Hi Jeremy,
Thanks for putting this together! I had a handful of inline comments, and also wanted to discuss the structure of the `RegisterMethod` class. The current setup requires the user to compute a reference and pass that into `getTransform()`. What about having the `ref` be a class variable of `RegisterMethod` and broadcasting that entire `RegisterMethod` object. So something like:

````
reg = Register('crosscorr')
reg.fitReference(images)
model = reg.fit(images)
````

where we would have:
```
def fit(images):
  ...
  reg_bc = images.rdd.context.broadcast(self)
  transformations = images.rdd.mapValues(lambda im: reg_bc.value.getTransform(im)).collectAsMap()
````

This would let the user store more complicated information in the `RegisterMethod` object that may be needed by `getTransform()`.


Minor comment -  how do you feel about changing `Register` to `Registration`?

File: python/thunder/imgprocessing/transformation.py (hunk 3 above)
Comment (freeman-lab) [inline python/thunder/imgprocessing/transformation.py]: Great catch, done.

**Comment by freeman-lab** [comment]
Thanks for the review @poolio ! Great suggestions. I really liked idea of making the reference a class attribute of the registration method, I've gone ahead and made that change. I made sure that the `fitReference` returns `self`, to make it easy to chain methods, as in:

```
model = Register('crosscorr').fitReference(images).fit(images)
newimages = Register('crosscorr').fitReference(images).run(images)
```

I also added a `setReference` method in case the user wants to provide a reference as an array directly (which seems like to happen sometimes). And I'm happy renaming `register` -> `registration`.

The issue of using numpy arrays for the delta, and whether and how to generalize the serialization handling, are actually related and require some thought. Let's discuss that one over coffee today =)

freeman-lab:
After a final offline discussion / review with @poolio and @broxtronix , going to merge this in.

PR #75, opened 2015-01-06
Camel casing

industrial-sloth:
Minor updates across the code base for better consistency with the style guide (docs/style_guide.rst). 

Most changes are to variable and method names, moving from an inconsistent mix of camel-casing, underscores, and all-lowercase to a more uniform use of camel casing throughout.


python/thunder/imgprocessing/register.py
```diff
@@ -162,27 +162,27 @@ def estimate(self, images, reference):
...
             else:
                 t = []
                 for z in arange(0, im.shape[2]):
-                    t.append(self.get_transform(self.filter(im[:, :, z]), ref.value[:, :, z]))
+                    t.append(self.getTransform(self.filter(im[:, :, z]), ref.value[:, :, z]))
             return t
 
         from thunder import Series
-        return Series(images.rdd.mapValues(lambda x: params(x, reference_bc)))
+        return Series(images.rdd.mapValues(lambda x: params(x, referenceBC)))
```
freeman-lab: Let's use `bcReference` and similar for broadcast variables

python/thunder/rdds/keys.py
```diff
@@ -60,67 +60,67 @@ def _indtosub_converter(dims, order='F', onebased=True):
...
 
-    def zerobased_prod(x_y):
-        x, y = x_y
+    def zerobasedProd(xy):
+        x, y = xy
         return x * y
 
-    def subtoind_inline_colmajor(k, dimprod, p_func):
-        return sum(map(p_func, zip(k[1:], dimprod))) + k[0]
+    def subToInd_InlineColmajor(k, dimProd_, prodFunc):
```
freeman-lab: Particular reason to use the `_` in this case? Seems like an exception to the rule, but maybe just too hard to read without?

python/thunder/rdds/matrices.py
```diff
@@ -110,7 +110,7 @@ def addInPlace(self, val1, val2):
                 return val1
 
         if method is "reduce":
-            return self.rdd.map(lambda (k, v): v).mapPartitions(matrixsum_iterator_self).sum()
+            return self.rdd.map(lambda (k, v): v).mapPartitions(matrixSum_Iterator_self).sum()
```
freeman-lab: Another use of `_`

python/thunder/rdds/series.py
```diff
@@ -541,26 +541,26 @@ def query(self, inds, var='inds', order='F', onebased=True):
...
-        for idx, indlist in enumerate(inds):
-            if len(indlist) > 0:
-                inds_set = set(asarray(indlist).flat)
-                inds_bc = self.rdd.context.broadcast(inds_set)
-                values[idx, :] = data.filterOnKeys(lambda k: k in inds_bc.value).values().mean()
-                keys[idx, :] = mean(map(lambda k: converter(k), indlist), axis=0)
+        for idx, indList in enumerate(inds):
+            if len(indList) > 0:
+                indsSet = set(asarray(indList).flat)
+                indsBC = self.rdd.context.broadcast(indsSet)
```
freeman-lab: Change to `bcInds`

freeman-lab:
Huge effort @industrial-sloth , nice job, it's looking good. Just left a couple small comments. These are significant changes, but it's important to ensure consistency from this point forward, and it will only get harder in the future. Luckily, most public classes and methods remain unchanged, the only user-facing changes should be to certain keyword argument names, which we will document in release notes.

python/thunder/rdds/keys.py
industrial-sloth: Hobgoblin of little minds, @freeman-lab. ;)

Yeah, here (and to some degree in the other remaining underscore) I thought that the underscore was useful in separating components of the name. subToIndInlineColmajor - are "subs" being converted to "IndInlineColmajors"? This version I think keeps it more clear that it's the "InlineColmajor" flavor of "subToInd"

python/thunder/rdds/matrices.py
industrial-sloth: Here I think one of the two underscores is actually useless, and I've gotten rid of it in the commit that I'm about to post, but I'm leaving it as "matrixSumIterator_self" (and "matrixSumIterator_other"), as these are the "self" and "other" varieties of a "matrixSumIterator".

industrial-sloth:
Thanks @freeman-lab, and thanks for the review! Have pushed a commit to resolve most of the issues you just raised - the couple remaining underscores I think do add to legibility, despite their being at variance with the style guidelines, and so I'd prefer to keep them (see inline comments). However I don't feel super strongly about this, and would be happy to blow them away entirely if you'd prefer.

freeman-lab:
LGTM! Merging it in.

PR #81, opened 2015-01-13
Nested directory reading

industrial-sloth:
This PR adds a 'recursive' keyword to all the image data input methods (basically, any time tifs or binary stacks are loaded from disk). If `recursive==True`, then the read methods will descend through a directory tree rooted at the passed `datapath` path, and try to load all files found in that tree with the specified file extension.

* This only works for reading from the local file system - specifying `recursive=True` when reading from S3 will result in an apologetic `NotImplementedError`.
* It also exposes an `ext` (extension) keyword argument on the `ThunderContext` object, allowing non-standard file extensions to be easily specified. Previously, to load flat binary image stacks with any extension other than "stack" (such as "bin"), one would have to manually create an ImagesLoader and specify the `ext` keyword there. Now .bin binary files can be read at the context level with just an additional keyword.
* This PR also makes file listing behavior a little more consistent between the local and parallel reader implementations, as well as between the BotoS3 and LocalFS readers. It is still not a perfect match, but now at least there are consistent `list()` methods on the local file readers, and `listFiles()` methods on the parallel readers. The main remaining differences are that the local (not parallel) readers are expected to be able to handle wildcard characters in the path specifier, while the parallel readers don't handle general wildcards, only allowing an optional extension to be specified. Also the BotoS3 readers do not support recursive listing of files, due mainly to the absence of an easy equivalent to `os.walk` in the boto API. 

(Just discovered a failing unit test as a result of this change - have pushed a commit to fix, but am not sure whether that will be picked up in the initial PR or no)

### DIFF
--- a/python/test/test_readers.py
+++ b/python/test/test_readers.py
@@ -1,7 +1,31 @@
+import os
+import shutil
+import tempfile
 import unittest
-from nose.tools import assert_raises
+from nose.tools import assert_equal, assert_raises
 
-from thunder.rdds.fileio.readers import LocalFSFileReader, FileNotFoundError
+from thunder.rdds.fileio.readers import LocalFSFileReader, LocalFSParallelReader, FileNotFoundError
+
+
+def touch_empty(filepath):
+    if not os.path.exists(filepath):
+        dirname, filename = os.path.split(filepath)
+        if not os.path.isdir(dirname):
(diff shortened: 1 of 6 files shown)


freeman-lab:
LGTM, thanks @industrial-sloth ! Great to have this option.

PR #82, opened 2015-01-14
add back pil_to_array function for use with old versions of PIL

industrial-sloth:
Old PIL does not load images correctly in Spark using the new-style `array(pilimg)` syntax, however the new pillow module appears to handle this correctly. Here fall back to matplotlib's old `pil_to_array` conversion if we detect that we've got PIL but not pillow installed.

Without this PR, loading tifs (at least single-page tifs, possibly others) with PIL 1.1.7 leads to results like the following:
```python
In [15]: data = tsc.loadImages('/path/to/tifs/', inputformat='tif-stack')
In [16]: data.first()[1].shape
Out[16]: (1, 1, 1)
In [17]: data.first()[1]
Out[17]: array([[[ <PIL.TiffImagePlugin.TiffImageFile image mode=I;16 size=512x512 at 0x3B02B00>]]], dtype=object)
```

### DIFF
--- a/python/thunder/rdds/fileio/imagesloader.py
+++ b/python/thunder/rdds/fileio/imagesloader.py
@@ -133,6 +133,16 @@ class ImagesLoader(object):
             Image = None
             raise ImportError("fromMultipageTif requires a successful 'from PIL import Image'; " +
                               "the PIL/pillow library appears to be missing or broken.", e)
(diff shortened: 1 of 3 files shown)


freeman-lab:
LGTM, thanks @industrial-sloth . Eventually it'd be best to eliminate our dependence on PIL/pillow entirely, but for now this is what we have to do.

PR #83, opened 2015-01-14
Windowed normalize optimization

jwittenbach:
Added a new option -- 'window-fast' -- to TimeSeries.normalize()
This is similar to the 'window' option, except that interpolation during the percentile computation is avoided by no using the numpy.percentile() method. Instead, the index within the window that is closest to the desired percentile is computed ahead of time. Then each window is sorted and this index is retrieved. Also updated the warning message for when widow size is set but user is not using one of the two windowed methods.

### DIFF
--- a/python/thunder/rdds/timeseries.py
+++ b/python/thunder/rdds/timeseries.py
@@ -304,11 +304,14 @@ class TimeSeries(Series):
                 left, right = (window/2 - 1, window/2)
 
             n = len(self.index)
-            ind_perc = round((window-1)*(perc/100)) 
+            ind_perc = round((window-1)*(perc/100.0)) 
             
             def basefunc(x):
                 left_pad = mean(x[:right])
-                right_pad = mean(x[-left:])
+                if left != 0:
+                    right_pad = mean(x[-left:])
+                else:
+                    right_pad = x[-1]
                 y = concatenate([ left_pad*ones(left), x, right_pad*ones(right) ])  
                 return asarray([sort(y[i-left:i+right+1])[ind_perc] for i in xrange(left,len(x)+left)])
 
--- a/python/thunder/rdds/timeseries.py
+++ b/python/thunder/rdds/timeseries.py
@@ -1,6 +1,6 @@
 from numpy import sqrt, pi, angle, fft, fix, zeros, roll, dot, mean, \
     array, size, diag, tile, ones, asarray, polyfit, polyval, arange, \
-    percentile, ceil, round, sort, ones
+    percentile, ceil, round, sort, ones, concatenate
 
 from thunder.rdds.series import Series
 from thunder.utils.common import loadmatvar, checkparams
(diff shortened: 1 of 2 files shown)


jwittenbach:
Just realized that I still need to write a new unit test for this method. Will do this and then push the changes.

python/thunder/rdds/timeseries.py
```diff
@@ -298,6 +297,21 @@ def normalize(self, baseline='percentile', window=None, perc=20):
...
+                left, right = (window/2 - 1, window/2)
+
+            n = len(self.index)
+            ind_perc = round((window-1)*(perc/100)) 
+            
+            def basefunc(x):
+                left_pad = mean(x[:right])
+                right_pad = mean(x[-left:])
+                y = concatenate([ left_pad*ones(left), x, right_pad*ones(right) ])  
+                return asarray([sort(y[i-left:i+right+1])[ind_perc] for i in xrange(left,len(x)+left)])
```
freeman-lab: Insert space after `left,`

python/thunder/rdds/timeseries.py
```diff
@@ -298,6 +297,21 @@ def normalize(self, baseline='percentile', window=None, perc=20):
             basefunc = lambda x: asarray([percentile(x[max(ix-left, 0):min(ix+right+1, n)], perc)
                                           for ix in arange(0, n)])
 
+        if method == 'window-fast':
+            if window%2:
```
freeman-lab: Insert white space around `%`

python/thunder/rdds/timeseries.py
```diff
@@ -298,6 +297,21 @@ def normalize(self, baseline='percentile', window=None, perc=20):
...
+                left, right = ((window-1)/2, (window-1)/2)
+            else:
+                left, right = (window/2 - 1, window/2)
+
+            n = len(self.index)
+            ind_perc = round((window-1)*(perc/100)) 
+            
+            def basefunc(x):
+                left_pad = mean(x[:right])
```
freeman-lab: Remove unscores by replacing with `leftPad` and `rightPad`. See the new style guide (to be added in 0.5.0): https://github.com/freeman-lab/thunder/blob/master/python/doc/style_guide.rst

freeman-lab:
Thanks @jwittenbach ! Code looks good except for some style / PEP convention things. You might want to enable PEP checking in PyCharm if you aren't already, that should catch things like missing spaces. And we need to figure out why the new tests are failing. It looks like it's the assertion on line 106, not yet obvious to me what the problem is.

jwittenbach:
I'm running the test locally and debugging that right now. I will push the new tests as well as the formatting as soon as I have it.

freeman-lab:
LGTM!

poolio:
Sorry for the late comment, but you may also be interested in `scipy.ndimage.filters.percentile_filter`  that uses a speedier Cython implementation. I can run some speed tests and submit a PR if you want.

freeman-lab:
@poolio thanks, awesome, didn't know about that. If it's faster would be great if you wanted to submit a PR with it (probably replacing the one we just added).

PR #84, opened 2015-01-15
Faster windowed normalization

poolio:
Speeds up the window-fast option just added in #83 using `scipy.ndimage.filters.percentile_filter`. 

I compared the windowed normalization techniques using a timeseries of length 50,000 and a window of 1,000:
````
window: 4.05s
window-fast-old: 2.64s
window-fast-new: 785ms
```
So we get a speedup of around 3x over the `window-fast` method in #83, and 5x over the original `window` method. These speedups were about the same over other window lengths (50, 500, 5000).


freeman-lab:
LGTM, thanks @poolio ! We did some testing on the cluster here and the speed up is great.

PR #87, opened 2015-01-19
Tif and tif stack

industrial-sloth:
This PR makes the 'tif' input format in the ThunderContext object resolve to the same execution path as 'tif-stack'. Mentions of 'tif-stack' have been removed from the context object docstrings in favor of encouraging the use of just plain 'tif', but 'tif-stack' remains a supported option. The old `ImagesLoader.fromTif` method, which only worked on single-page tifs, has been removed, and the various `fromMultipageTif` methods have been renamed to just `fromTif`. 

From the user's perspective, the only noticeable behavior change should be that single-page tif files will now appear as three-dimensional images (x by y) instead of 3d (x by y by 1) as they came out of the previous `loadMultipageTif` method. SeriesLoader has also been changed to produce x by y (instead of x by y by 1) Series data from single-page tifs for consistency with the new ImagesLoader output.


industrial-sloth:
Hmmm, unit tests are passing on my machine, not immediately sure what's up here...

industrial-sloth:
Ah. pil_to_array (which is only used on machines with PIL but not pillow) complains, rightly, about RGBA images. The test images in `resources/singlelayer_tif` and `resources/singlelayer_png` are RGBA. I'll convert these over to luminance images.

freeman-lab:
Thanks @industrial-sloth , LGTM, merging it in.

PR #89, opened 2015-01-22
support multiple time points per image file

industrial-sloth:
This PR adds an `nplanes` option to the main Images-loading methods. If `nplanes` is specified, then a single input file will be interpreted as containing multiple image volumes, each with size `nplanes` in its final dimension. For instance, a single binary stack file loaded with arguments `dims=(x, y, 8), nplanes=2` would turn into 4 separate records in an Images RDD, each with size `(x, y, 2)`. In general, images that are loaded with `z` planes and a positive `nplanes` argument will result in `z / nplanes` time points, each with `nplanes` planes.


industrial-sloth:
Meh. This has some overlap with PR #87, which changes the expected dimensions of images with only a single z-plane from `(x, y, 1)` to just `(x, y)`. I've written this code anticipating this new behavior, but forgot to update all the tests accordingly. Will pull in the relevant changes from PR #87 and put those in a new commit here as well to get the tests passing again.

freeman-lab:
@industrial-sloth you were right, merging the other `tif` PR created a conflict with this one.

industrial-sloth:
Whoops, missed changing some function names in the merge, fixing.

python/thunder/rdds/fileio/imagesloader.py
```diff
@@ -75,34 +76,82 @@ def fromStack(self, dataPath, dims, dtype='int16', ext='stack', startIdx=None, s
             If true, will recursively descend directories rooted at datapath, loading all files in the tree that
             have an extension matching 'ext'. Recursive loading is currently only implemented for local filesystems
             (not s3).
+
+        nplanes: positive integer, default None
+            If passed, will cause a single binary stack file to be subdivided into multiple time points. Every
```
freeman-lab: Let's avoid the explicit references to `time`. I just checked and we're generally good about this, though in fixing this would be great if you caught other users. There's certainly a notion of the images being /ordered/ both here and elsewhere, but that ordering need not relate to time specifically.

python/thunder/rdds/fileio/imagesloader.py
```diff
@@ -120,27 +169,44 @@ def fromTif(self, dataPath, ext='tif', startIdx=None, stopIdx=None, recursive=Fa
             from thunder.utils.common import pil_to_array
             conversionFcn = pil_to_array  # use our modified version of matplotlib's pil_to_array
 
-        def multitifReader(buf):
+        if nplanes is not None and nplanes <= 0:
```
freeman-lab: Is it well defined in the case where `nplanes=1`?

python/thunder/utils/context.py
```diff
@@ -145,6 +145,11 @@ def loadImages(self, dataPath, dims=None, inputFormat='stack', ext=None, dtype='
             have an appropriate extension. Recursive loading is currently only implemented for local filesystems
             (not s3).
 
+        nplanes: positive integer, default None
+            If passed, will cause a single image file to be subdivided into multiple time points. Every `nplanes` image
```
freeman-lab: Here also the point about referring to time explicitly.

python/thunder/utils/context.py
```diff
@@ -376,6 +393,12 @@ def convertImagesToSeries(self, dataPath, outputDirPath, dims=None, inputFormat=
             If true, will recursively descend directories rooted at dataPath, loading all files in the tree that
             have an appropriate extension. Recursive loading is currently only implemented for local filesystems
             (not s3), and only with shuffle=True.
+
+        nplanes: positive integer, default None
+            If passed, will cause a single image file to be subdivided into multiple time points. Every `nplanes` image
```
freeman-lab: And again =)

freeman-lab:
@industrial-sloth this is looking good, left a couple minor notes. One other question: how does the ordering work? Am I right in assuming that, given three (X,Y,Z) image files, call them 1, 2, 3, and setting nplanes=Z/2, you would end up with an ordering of 1_1, 1_2, 2_1, 2_2, 3_1, 3_2, where 1_1 and 1_2 are the first and second chunks of the first image, etc.. In other words, it'll be ordered first by the image files, then by the chunks. That seems like what we want, just making sure.

python/thunder/rdds/fileio/imagesloader.py
industrial-sloth: Yes? `nplanes=1` would indicate that you've actually got a 2d image, where each separate page in the multipage tif corresponds to a different image of the same 2d plane (I am struggling to avoid saying "time point" here :) ). There's a test, `test_fromMultipleMultiTimepointTifs`, that specifies `nplanes=1` and which behaves as expected. Before you ask, yes, the shape in this case is 2d (x,y), and not 3d (x,y,1). :)

But I'm not sure if the question I'm answering is the one you're actually asking? Let me know if I'm off track here.

industrial-sloth:
@freeman-lab yes, that's the behavior. For concreteness, if we had N 10-page tif files and set `nplanes=5`, the first five pages of the first tif file would end up as an Images record with key 0. The second five pages of the first file would be the record with key 1, the first five pages of the second file would be record 2, and so on. (I'm pretty sure we're saying the same thing here, but just wanted to check. :)

industrial-sloth:
Ok, I've changed the docstrings to talk about generic "records" rather than "time points". Looks like there's a merge conflict to resolve, will fix that momentarily.

industrial-sloth:
Have added checks that `nplanes` does evenly divide the number of pages / planes in an image file to the tif and binary stack readers. A slight oddity - for the tif reader, it looks like the most natural home for this check is on the executors, not the driver, since the driver doesn't know how many pages a tif file has got. At the moment then the *executors* throw `ValueError` if `nplanes` doesn't divide the total page count evenly. This leads to a somewhat weird looking stack dump, but is still interpretable? I think? 

The flat binary stack file case is just fine, since there the driver knows what the dimensions are ahead of time and can do the check before launching jobs.

freeman-lab:
Thanks @industrial-sloth , agreed that doing the check on the executors for the `tifs` is kind of the only option for now, short of some convoluted process by which we load one to check, and then proceed if it's ok. I think this is fine, will merge this version in.

PR #93, opened 2015-01-22
Series index setter

jwittenbach:
All the setter does is type-check and then set, so I didn't think a unit test was terribly necessary. But if we want one, I can add it. I'll just how to figure out how to assert that the correct exception is being raised...

### DIFF
--- a/python/thunder/rdds/series.py
+++ b/python/thunder/rdds/series.py
@@ -62,15 +62,16 @@ class Series(Data):
         try:
             value[0]
         except:
-            raise ValueError("Index must be a non-empty indexable object")
+            raise AttributeError("Index must be a non-empty indexable object")
         if self._index is None:
             self.populateParametersFromFirstRecord()
         try:
-            lenMatch = len(value) == len(self._index)
+            lenValue = len(value)
+            lenSelf = len(self._index)
+            if lenValue != lenSelf:
+                raise ValueError("Length of new index ({1}) must match length of original index ({2})".format(lenValue, lenSelf))
         except:
-            raise ValueError("Index must be an object with a length")
-        if not lenMatch:
-            raise ValueError("New index length must match that of the current index")
+            raise TypeError("Index must be an object with a length")
         self._index = value
 
     @property
--- a/python/thunder/rdds/series.py
+++ b/python/thunder/rdds/series.py
@@ -63,9 +63,10 @@ class Series(Data):
             value[0]
         except:
             raise ValueError("Index must be a non-empty indexable object")
-        if not self._index is None:
-            if len(value) != len(self._index):
-                raise ValueError("New index length must match that of the current index")
+        if self._index is None:
+            self.populateParametersFromFirstRecord()
+        if len(value) != len(self._index):
+            raise ValueError("New index length must match that of the current index")
         self._index = value
 
     @property
--- a/python/thunder/rdds/series.py
+++ b/python/thunder/rdds/series.py
@@ -63,13 +63,11 @@ class Series(Data):
             value[0]
         except:
             raise AttributeError("Index must be a non-empty indexable object")
-        if self._index is None:
-            self.populateParametersFromFirstRecord()
         try:
             lenValue = len(value)
         except:
             raise TypeError("Index must be an object with a length")
-        lenSelf = len(self._index)
+        lenSelf = len(self.index)
         if lenValue != lenSelf:
             raise ValueError("Length of new index ({0}) must match length of original index ({1})".format(lenValue, lenSelf))
         self._index = value
--- a/python/thunder/rdds/series.py
+++ b/python/thunder/rdds/series.py
@@ -67,11 +67,11 @@ class Series(Data):
             self.populateParametersFromFirstRecord()
         try:
             lenValue = len(value)
-            lenSelf = len(self._index)
-            if lenValue != lenSelf:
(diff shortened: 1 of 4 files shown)


industrial-sloth:
Good deal. Let's do get a unit test or two on this, just to ensure in the future that the exceptions we expect to get raised actually stay raised. Check out `assert_raises` in `nose.tools` - there's a usage of it in `test_images.py`, probably other places too.

python/thunder/rdds/series.py
```diff
@@ -56,6 +56,22 @@ def index(self):
...
+        try:
+            value[0]
+        except:
+            raise ValueError("Index must be a non-empty indexable object")
+        if self._index is None:
+            self.populateParametersFromFirstRecord()
+        try:
+            lenMatch = len(value) == len(self._index)
+        except:
+            raise ValueError("Index must be an object with a length")
```
industrial-sloth: Here I think the traditional exception would be `AttributeError`? I like the descriptive message though, so it's worth catching and raising your own exception here like you're doing.

python/thunder/rdds/series.py
```diff
@@ -56,6 +56,22 @@ def index(self):
...
+        if not lenMatch:
+            raise ValueError("New index length must match that of the current index")
```
industrial-sloth: I can imagine somebody getting really frustrated here if they think the length of the new index they're passing *does* already match the length of the current index. How about adding printouts of the length of the new index and the existing index to the exception text? something like `"New index length (%d) ...current index (%d)" % (len(value), len(self._index))`

Also I don't think the `lenMatch` var is really needed here, Could do:
```python
try:
  if len(val) != len(self._index):
    raise blah blah
except:
  raise another blah
```

python/thunder/rdds/series.py
jwittenbach: Is it standard to use the built-in Python exception classes? Or should I define my own?

industrial-sloth: Using the existing classes is just fine if they accurately describe the error that you're seeing. I'm probably guilty of overusing `ValueError` myself (but "inappropriate value" describes, like, everything! https://docs.python.org/2/library/exceptions.html)

The one custom exception I can think of offhand in Thunder is a `FileNotFoundException` defined in reader.py. It's a little silly that this doesn't already exist I think - and depending on the exact context, having a missing file could result in either `IOError` or `OSError`. Personally, I haven't run into that many cases where I thought having an exception class of my very own would be all that useful. But don't let this stop you. :)

jwittenbach: I was actually just looking at that webpage :smile_cat:

python/thunder/rdds/series.py
jwittenbach: I might be wrong, but I think this won't work. In the even that the if-statement fails, it will raise "blah blah", which the except will then catch, and raise "another blah"...so the first exception will never be raised. At least that's what my initial testing is telling me.

jwittenbach: I do like the more descriptive error message though!

industrial-sloth: Oh sorry - that of course should have been:
```python
try:
  if len(val) != len(self._index):
    raise ValueError(blah blah)
except TypeError:
  raise TypeErrorOrSomething(another more descriptive blah)
```
Or you could just let the initial TypeError propagate on through. But whatever, really not a big deal. :)

python/thunder/rdds/series.py
```diff
@@ -56,6 +56,23 @@ def index(self):
...
             self.populateParamsFromFirstRecord()
         return self._index
+        
+    @index.setter
+    def index(self, value):
+        try:
+            value[0]
+        except:
+            raise AttributeError("Index must be a non-empty indexable object")
+        if self._index is None:
```
freeman-lab: This shouldn't be necessary if, in line 72, you use `self.index` instead of `self._index`, because the automatic population from the first record will happen there.

jwittenbach:
I think this most recent edition should address everything we've talked about so far.

python/thunder/rdds/series.py
```diff
@@ -307,7 +327,8 @@ def seriesPercentile(self, q):
           q: a floating point number between 0 and 100 inclusive.
         """
         rdd = self.rdd.mapValues(lambda x: percentile(x, q))
-        return self._constructor(rdd, index='percentile').__finalize__(self, noPropagate=('_dtype',))
+        print rdd.first()
```
freeman-lab: Think you left in a stray `print` here.

python/thunder/rdds/series.py
```diff
@@ -43,7 +43,9 @@ class Series(Data):
 
     def __init__(self, rdd, index=None, dims=None, dtype=None):
         super(Series, self).__init__(rdd, dtype=dtype)
-        self._index = index
+        self._index = None
+        if not index is None:
```
freeman-lab: Probably replace with `if index is not None:` as it's less ambigious, see http://stackoverflow.com/questions/2710940/python-if-x-is-not-none-or-if-not-x-is-none for discussion

jwittenbach:
Thanks for catching those. I also cleaned up the code in the setter a little bit and added comments to explain the logic a little.

freeman-lab:
Thanks @jwittenbach , LGTM, seem fine to you @industrial-sloth ?

industrial-sloth:
Whoa, sorry, forgot this was waiting on me. Yep, looking good, merging.

PR #98, opened 2015-02-05
Data getter

industrial-sloth:
This adds support for `get`ting individual records or sets of records from `Data` subtypes. 

Three methods are added, together with `__getitem__` (brackets notation) as a syntactic convenience. The methods are:

* `get`: Returns a single record value, specified by record key (example: `series.get((12,24,36))`, returns series value at (12, 24, 36))
* `getAll`: Returns multiple record values, one for each key passed (example: `images.getAll([2, 4, 6])` returns a sequence of image arrays from time points 2, 4, and 6.
* `getRange`: returns multiple record values specified by slices (example: `series.getRange([slice(1), slice(2,3)]` returns the single series value with key (0, 2).
* brackets notation delegates either to `get` or to `getRange` depending on whether a single key or a range of keys are passed (example: `series[:3,2:3]` returns values of keys (0,2), (1,2), (2,2)).


industrial-sloth:
Hmmm, no sooner do I create the PR than I realize what I'm missing - need to add support for a mixture of slices and individual indices in getRange (and the brackets accessor). Coming right up... also apparently I need to resolve a merge conflict...

python/thunder/rdds/data.py
```diff
@@ -111,6 +111,129 @@ def take(self, *args, **kwargs):
...
+                            return False
+                    elif not (slise.stop > subkey >= slise.start):
+                        return False
+                else:  # not a slice
+                    if subkey != slise:
+                        return False
+            return True
+
+        if not hasattr(sliceOrSlices, '__len__'):
+            # make my func the...
```
freeman-lab: Don't understand this note =)

python/test/test_series.py
```diff
@@ -1,6 +1,6 @@
 from numpy import allclose, amax, arange, array, array_equal
```
freeman-lab: Any reason to split into two test modules? It's a `Data` method, so my first impulse would have been to put all these tests, with the full variety of keys, inside `test_data.py`. The conventions we use for integer keys on `Images` and tuple keys on `Series`, while common, are not strictly enforced, so seems a little weird to split the tests up based on that.

python/thunder/rdds/data.py
```diff
@@ -111,6 +111,129 @@ def take(self, *args, **kwargs):
...
         return self.rdd.take(*args, **kwargs)
 
+    def get(self, key):
+        """Returns a single value matching the passed key, or None if no matching keys found
+
+        If multiple records are found with keys matching the passed key, a sequence of all matching
+        values will be returned. (This is not expected as a normal occurance, but could happen with
+        some user-created rdds.)
+        """
+        filteredVals = self.rdd.filter(lambda (k, v): k == key).values().collect()
```
freeman-lab: Could also use our `filterOnKeys` method, would simplify the lambda and avoid dereferencing the rdd, right?

python/thunder/rdds/data.py
```diff
@@ -111,6 +111,129 @@ def take(self, *args, **kwargs):
...
+        """Returns a sequence of values corresponding to the passed sequence of keys.
+
+        The return value will be a sequence equal in length to the passed keys, with each
+        value in the returned sequence corresponding to the key at the same position in the passed
+        keys sequence. If no value is found for a given key, the corresponding sequence element will be None.
+        If multiple values are found, the corresponding sequence element will be a sequence containing all
+        matching values.
+        """
+        keySet = frozenset(keys)
+        filteredRecs = self.rdd.filter(lambda (k, _): k in keySet).collect()
```
freeman-lab: Same suggestion as above about using `filterOnKeys`

python/thunder/rdds/data.py
```diff
@@ -111,6 +111,129 @@ def take(self, *args, **kwargs):
...
+                if hasattr(slise, 'step') and slise.step is not None:
+                    raise ValueError("'step' slice attribute is not supported in getRange, got step: %d" %
+                                     slise.step)
+
+        filteredRecs = self.rdd.filter(pFunc).collect()
+        # default sort of tuples is by first item, which happens to be what we want
+        return sorted(filteredRecs)
+
+    def __getitem__(self, item):
+        # should raise exception here when no matching items found
```
freeman-lab: Good, I like this.

freeman-lab:
@industrial-sloth this is really nice! It's looking great, some minor comments, main one is about possibly reorganizing the tests.

python/thunder/rdds/data.py
```diff
@@ -111,6 +111,129 @@ def take(self, *args, **kwargs):
...
+        """
+        filteredVals = self.rdd.filter(lambda (k, v): k == key).values().collect()
+        if len(filteredVals) == 1:
+            return filteredVals[0]
+        elif not filteredVals:
+            return None
+        else:
+            return filteredVals
+
+    def getAll(self, keys):
```
freeman-lab: Rename to `getMany` to avoid confusion with `collect`

industrial-sloth:
Ok, have updated this PR based on offline conversations with @freeman-lab. 
Changes:
* tests are now in a new test_data.py file rather than test_images and test_series
* the various `get` methods now call `first` in order to perform type checking against actual RDD keys, throwing exceptions early if the queried key doesn't match the arity of the actual keys
* `getAll` has been renamed to `getMany`
* the brackets syntax for the getter methods now returns None or [] when queries fail to return any values, rather than throwing KeyError or IndexError.

freeman-lab:
Thanks @industrial-sloth ! LGTM, merging it in.

PR #95, opened 2015-01-27
Aggregate regions

industrial-sloth:
This PR adds two closely-related methods to Series objects, `meanOfRegion` and `meanByRegions`. These calculate the average time series within a set or sets of Series records specified by their keys.

PR #99, opened 2015-02-05
Son of Aggregate regions: this is really aggregating

industrial-sloth:
This PR expands on the functionality introduced in PR #95, adding support for specifying regions via an array mask on the Series mean of/by region(s) methods, and adding a `meanByRegions` method on Images. 

The new images method currently returns Images where each record value is a 1 x N array, with N being the number of regions specified. This can then be converted into time series values via the usual `images.toSeries().collect()` dance, which should run very quickly since after the aggregation into regions there should be almost no data left. :) The Images meanByRegions method supports region specification either by a mask array or by passing a sequence of sequences of indices, as in the equivalent Series method.

### DIFF
--- a/python/thunder/rdds/images.py
+++ b/python/thunder/rdds/images.py
@@ -459,6 +459,7 @@ class Images(Data):
             means = array([mean(ary[maskIdxs]) for maskIdxs in maskIdxsSeq], dtype=ary.dtype).reshape((1, -1))
             return key, means
 
+        nregions = -1
         # argument type checking
         if isinstance(selection, ndarray):
             # passed a numpy array mask
@@ -472,10 +473,13 @@ class Images(Data):
             if selection.dtype.kind in ('i', 'u'):
                 # integer or unsigned int mask
                 selectFcn = meanByIntMask
-                bcUnique = ctx.broadcast(unique(selection))
+                uniq = unique(selection)
+                nregions = len(uniq) - 1 if 0 in uniq else len(uniq)  # 0 doesn't turn into a region
+                bcUnique = ctx.broadcast(uniq)
                 bcSelection = ctx.broadcast(selection)
             else:
                 selectFcn = meanByMaskIndices
+                nregions = 1
                 bcUnique = None
                 bcSelection = ctx.broadcast((selection.nonzero(), ))
         else:
@@ -503,11 +507,12 @@ class Images(Data):
                                          (idxMin, idxMax, imgDimMax))
                     regionSelect.append(dimIdxAry)
                 regionSelections.append(regionSelect)
+            nregions = len(regionSelections)
             bcUnique = None
             bcSelection = ctx.broadcast(regionSelections)
 
         data = self.rdd.map(selectFcn)
-        return self._constructor(data).__finalize__(self)
+        return self._constructor(data, dims=(1, nregions)).__finalize__(self)
 
     def planes(self, startidz, stopidz):
         """
--- a/python/thunder/rdds/images.py
+++ b/python/thunder/rdds/images.py
@@ -416,6 +416,65 @@ class Images(Data):
 
         return self._constructor(newrdd, dims=newdims).__finalize__(self)
 
+    def meanByRegions(self, mask):
+        """Collapses images into one or more spatial mean values as specified by the passed mask array.
+
+        The passed mask must be a numpy ndarray of the same shape as the individual arrays in this
+        Images object. If the mask array is of integer or unsigned integer type, one mean value will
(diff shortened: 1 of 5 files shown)


python/thunder/rdds/images.py
```diff
@@ -416,6 +416,85 @@ def crop(self, minbound, maxbound):
 
         return self._constructor(newrdd, dims=newdims).__finalize__(self)
 
+    def meanByRegions(self, selection):
+        """Collapses images into one or more spatial mean values as specified by the passed mask array or sequence of
```
freeman-lab: Suggested rewording: `Reduces images to one or more spatially averaged values using the given mask array or sequence of indices`

python/thunder/rdds/images.py
```diff
@@ -416,6 +416,85 @@ def crop(self, minbound, maxbound):
...
+        indicies.
+
+        A passed mask must be a numpy ndarray of the same shape as the individual arrays in this
+        Images object. If the mask array is of integer or unsigned integer type, one mean value will
+        be calculated for each unique nonzero value in the passed mask. (That is, all pixels with a
```
freeman-lab: Very nice!

python/thunder/rdds/series.py
```diff
@@ -582,18 +582,80 @@ def query(self, inds, var='inds', order='F', isOneBased=True):
...
+            else:
+                return keys
+
+    def meanOfRegion(self, selection):
+        """Takes the mean of Series values within a single region specified by the passed mask or keys.
+
+        The region for which to take the mean may be specified either by a mask array, or directly by
+        Series keys. If an ndarray is passed as `selection`, then the mean will be taken across all series
+        records corresponding to nonzero elements of the passed mask. (The passed ndarray must have the
+        same shape as series.dims.count, otherwise a ValueError will be thrown.)
```
freeman-lab: Maybe add here and in `meanByRegion` that using a mask is only well-defined when the `Series` records are indexed by contiguous ranges in all dimensions? I'm fine not checking for that, but it would be good to warn the user.

python/thunder/rdds/series.py
```diff
@@ -582,18 +582,80 @@ def query(self, inds, var='inds', order='F', isOneBased=True):
...
+        Returns
+        -------
+        sequence of subscripted indices if returnNested is false,
+        sequence of sequence of subscripted indices if returnNested is true
+        """
+        # argument type checking
+        if not isinstance(mask, ndarray):
+            raise ValueError("Mask should be numpy ndarray, got: '%s'" % str(type(mask)))
+        # check for matching shapes only if we already know our own shape; don't trigger action otherwise
+        if self._dims:
```
freeman-lab: This is a bit weird. I see the rationale for not triggering. However, if the user violates this assumption, and we don't check, it may actually run to completion with no errors (for example, it may just not find the full set of regions the user expected), right? Normally, I'm fine avoiding an expensive pre-computation if we know the user will get an error on the workers down the road. But here that might not be the case. So maybe we always want to check ahead of time? Unfortunately, there's no way to also check that the keys are contiguous, which is another implicit assumption of using the mask...

freeman-lab:
@industrial-sloth code looks really clean, my main concern is the checking we do / assumptions we make when using a mask on Series data. It gets into the tricky issue of what assumptions we're making about the keys (if any), and how to check for them (or not). Might be worth discussing offline.

python/thunder/rdds/series.py
industrial-sloth: I saw this comment after pushing some changes and was briefly worried that I'd forgotten to address it - but no, for my own future reference: after offline discussion with @freeman-lab, we decided to address this concern by adding additional checks to the series methods, which now by default will throw an error if specified keys are not found in the Series object.

industrial-sloth:
OK, have pushed changes to add additional range checking to the passed selection objects in both the Series and Images versions of these methods, and additional checks in the Series version to ensure that the regions we generate are the result of taking the mean over the expected number of records. This last is to address the concern that the method might be called on a "sparse" Series object, where the keys no longer make up a contiguous range. I think this PR should now be ready for an additional round of review.

freeman-lab:
This is close, my main suggestions (see comments) were to simplify the options / settings for the error checking in the `Series` case, and an idea for avoiding the computation when doing the check in `Series.meanByRegion`.

industrial-sloth:
Alright, I am loving the simplified error check in `Series.meanByRegion`. I think these commits should address all the comments, so this may be ready to go?

freeman-lab:
LGTM!

PR #102, opened 2015-02-10
Changed bracket behavior for Data

Andrewosh:
The __getitem__ method on Data objects now raises an KeyError if a value is not found for the given
key. Previously it returned None.  The Data unit tests have been updated accordingly.


freeman-lab:
@Andrewosh thanks! Looks like a couple tests are failing, check out `thunder/python/test/test_data.py`, lines 104 and 254. Probably because they are now getting an error and expected `None`.

PR #107, opened 2015-02-13
Series manipulations by index value

jwittenbach:
Adds methods to `Series` to allow for selection, filtering, and aggregation of data in `Series` records based in the `Series` index.

One of the main targets of such manipulations are "multi-indexed" arrays -- arrays where the index is a list of lists, each sublist being of the same length. In this case, each element of the index will have values at multiple "levels". These levels can be useful for describing additional structure on the series by allowing multiple sets of labels on the data.

---

#### Functions

*  `seriesSelectByIndex`
Allows the user to select one or multiple levels of the index as well as one or several values for each of these levels. Returns a new `Series` where only the subset of values in each record where the index matches the values at the given levels are retained. Has an option to filter instead, only returning those values that do NOT match (this could easily be moved to a separate function if desired). Also has an option to remove any levels from the multi-index where only a single value was selected; this in the process of keeping unique indices if the user desires this.

*  `seriesAggregateByIndex`
Allows the user to select one of multiple levels of the index. The user also supplies an aggregating function. Data in each record is first grouped based on the values of the index at these levels, one group per unique set of values. Returns a new `Series` where the values in the records are the results of applying the aggregating function to each group. Ideally this function would be many-to-one, though this is not demanded at the moment.

* Convenience functions
Also included are a set of convenience functions that apply `seriesAggregateByIndex` with predefined functions.

---

#### Notes

* All functions default to only using the first level if none are supplied. This allows these functions to do the sensible thing when the `Series` is not multi-indexed and the user need not even think about levels.
* When only a single level or value is required, the user can pass in values instead of lists of values for convenience
* In the future, it might be nice to add support for "naming" the levels/values in the index. Right now, levels must be given as integers.


python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+
+        if index is None:
+            index = self.index
+        
+        try:
+            dims = len(array(index).shape)
+            if dims == 1:
+                index = array(index, ndmin=2).T
+        except:
+            raise TypeError('for multi-index functionality: index must be convertible to a numpy ndarray')
```
freeman-lab: Start error message with capital letter

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+
+        try:
+            index = index[:, level]
+        except:
+            raise ValueError("levels must be indices into individual elements of the index")
```
freeman-lab: Start error message with capital letter

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+            for i in xrange(len(val)):
+                try:
+                    val[i][0]
+                except:
+                    val[i] = [val[i]]
+                if squeeze and not filter and len(val[i]) == 1:
+                    remove.append(level[i])
+                                
+        if len(level) != len(val):
+            raise ValueError("list of levels must be of same length as list of corresponding values")
```
freeman-lab: Same note about capitalization

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        masks, ind = self._makeMasks(index=self.index, level=level)
+        nmasks = len(masks)
+        masks = array([masks[x] for x in xrange(nmasks) if tuple(ind[x]) in selected])
+
+        finalMask = masks.any(axis=0)
+        if filter:
+            finalMask = logical_not(finalMask)
+        bcMask = self.rdd.ctx.broadcast(finalMask)
+        
+        rdd = self.rdd.mapValues(lambda v: v[bcMask.value])
```
freeman-lab: Call this newrdd, we've used that convention elsewhere

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        if squeeze:
+            indFinal = delete(indFinal, remove, axis=1)
+
+        if len(indFinal[0]) == 1:
+            indFinal = ravel(indFinal)
+
+        elif len(indFinal[1]) == 0:
+            indFinal = arange(sum(finalMask))
+
+        return Series(rdd, index=indFinal).__finalize__(self, noPropagate=('_index',))
```
freeman-lab: The `noPropagate` shouldn't be necessary if an index is being passed during construction, but double check that

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        to to each of these groups separately. If this function is many-to-one, the result
+        can be recast as a Series indexed by the unique index values used for grouping.
+        """
+
+        if type(level) is int:
+            level = [level]
+
+        masks, ind = self._makeMasks(index=self.index, level=level)
+        bcMasks = self.rdd.ctx.broadcast(masks)
+        bcNum = self.rdd.ctx.broadcast(len(masks))
```
freeman-lab: If this is just a single number definitely no need to broadcast, just refer to it inside the function.

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        def f(data):
+            return array([function(data[bcMasks.value[x]]) for x in xrange(bcNum.value)])
+        rdd = self.rdd.mapValues(f)
```
freeman-lab: Again, let's use newrdd by convention

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        """
+
+        if type(level) is int:
+            level = [level]
+
+        masks, ind = self._makeMasks(index=self.index, level=level)
+        bcMasks = self.rdd.ctx.broadcast(masks)
+        bcNum = self.rdd.ctx.broadcast(len(masks))
+        def f(data):
+            return array([function(data[bcMasks.value[x]]) for x in xrange(bcNum.value)])
```
freeman-lab: This might be short enough to do as a lambda, rather than define a new function

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        rdd = self.rdd.mapValues(f)
+        index = array(ind)
+        if len(index[0]) == 1:
+            index = ravel(index)
+        return Series(rdd, index=index).__finalize__(self, noPropagate=('_dtype', '_index'))
```
freeman-lab: See note above about propagating indices when provided in the constructor

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+
+    def selectByIndex(self, level=0, val=None, squeeze=False, filter=False):
```
freeman-lab: I think a brief description of a multilevel index would be useful, probably both here and in the `seriesAggregateByIndex` methods. Don't need it in the private functions.

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        """
+        Select or filter elements of the Series by index values (across levels, if multi-index)
+
+        At each of the given levels, only values with a multi-index that matches one of the given
+        values will be retained for the returned Series. If filtering, this is inverted and only 
+        those elements will be ddropped.
+
+        Parameters:
+        -----------
+        level: An integer or an list of integers specificying which levels in the multilevel index to use when
```
freeman-lab: For parameter formatting in docstrings, the first line should be `argname : type`, as in `level: integer or list`, as well as a note about whether it's optional and what the default is, like `val: boolean, optional, default = False`. Then the second line should be a description like what you have here, but without restating the type to avoid redundancy.

python/thunder/rdds/series.py
```diff
@@ -883,6 +885,229 @@ def toSpatialSeries(self):
...
+        Aggregrate the data in each record, grouping by a multilevel index
+        
+        For each unique value of the index (across levels, if multi-index) an aggregating function is applied to the
+        list of all values with that index (ordered by location within the series, for the case where the function
+        is not transitive). This is done on a record-by-record basis. Returns a new Series with the results of this
+        operation, indexed by the unique indices used for the grouping.
+
+        Parameters:
+        -----------
+        level: An integer or an list of integers specifying which levels in the multilevel index to use when
```
freeman-lab: See other note about docstring formatting

python/test/test_series.py
```diff
@@ -249,6 +249,83 @@ def setIndex(data, idx):
...
+            [0, 0, 1, 1, 1, 1, 0, 0, 1, 1, 1, 1],
+            [0, 1, 0, 1, 2, 3, 0, 1, 0, 1, 2, 3]
+        ]
+        data.index = array(index).T
+        
+        result = data.seriesAggregateByIndex(level=[0,1], function=sum)
+        assert_true(array_equal(result.values().first(), array([1, 14, 13, 38])))
+        assert_true(array_equal(result.index, array([[0, 0], [0, 1], [1, 0], [1, 1]])))
+
+    def test_seriesStatByIndex(self):
```
freeman-lab: Include tests here that use the multilevel indexing?

python/thunder/rdds/series.py
freeman-lab: And especially a little more clarity on the difference between single level and multi-level indices.

jwittenbach:
I think address all of the comments. Let me know what you think of the doc-string format now.

python/thunder/rdds/series.py
```diff
@@ -892,6 +895,248 @@ def toSpatialSeries(self):
...
+        nMasks = len(masks)
+        newrdd = self.rdd.mapValues(lambda v: [array(function(v[bcMasks.value[x]])) for x in xrange(nMasks)])
+        index = array(ind)
+        if len(index[0]) == 1:
+            index = ravel(index)
+        return Series(newrdd, index=index).__finalize__(self, noPropagate=('_dtype',))
+
+    def selectByIndex(self, val, level=0, squeeze=False, filter=False):
+        """
+        Select or filter elements of the Series by index values (across levels, if multi-index)
```
freeman-lab: nit: add a period at the end

python/thunder/rdds/series.py
```diff
@@ -892,6 +895,248 @@ def toSpatialSeries(self):
         from thunder.rdds.spatialseries import SpatialSeries
         return SpatialSeries(self.rdd).__finalize__(self)
 
+    def _makeMasks(self, index=None, level=0):
+        """
+        Internal function for generating masks for selecting values based on multi-index values
```
freeman-lab: nit: add a period at the end

jwittenbach:
Turns out the problem was that Data.apply has default behavior that does not keep the propagate the index to the result (Is this new?). Since my aggregation uses apply, this was an issue. Should be fixed now!

freeman-lab:
@jwittenbach great catch, yes, that behavior changed because, in general, we have no idea what `apply` does, so assuming it might change the index seemed (and force a new one to be computed) seemed safer than assuming it left the index unchanged.

freeman-lab:
Going to merge this in, thanks for the contribution @jwittenbach !

PR #112, opened 2015-02-18
Modified sorting

freeman-lab:
This PR reorganizes support for sorting outputs, 

A couple key changes / hopefully improvements:

* The primary sorting behavior is relegated to a `Data` method `sortByKey()` that calls the underlying Spark RDD functions. This will make it easier to maintain / modify the sorting behavior and use it elsewhere.
* By using this function, sorting during `Series.pack()` is now done with RDD functions, which avoids an error by which collecting the keys and values separately (for driver-side sort) can yield different orderings.
* The `collect`, `collectAsArray`, `collectValuesAsArray`, and `collectKeysAsArray` methods all now take `sorting` as an input argument. This has previously been an option on `pack`, but it is useful as an option on these functions for non-contiguous data (as reported by @jwittenbach ).

I have not yet done performance testing on driver-side sorting versus `sortByKey()` but especially with improvements to shuffling in Spark 1.2 I expect it to be comparable if not better (it previously was the case that performance sorting smaller results especially was much better on the driver). If we do want to revert, with the new refactoring, we'll only need to make the switch in one place.

For sorting behavior consistent with other functionality / conventions in Thunder (fastest key changes first), I do reverse the ordering of the keys before (and after) sorting. The one consequence is that the sorting methods will break for integer keys. Keys are *supposed* to always be tuples, but we also don't / can't enforce that. Perhaps the sorting method should check to make sure the keys are iterable before sorting and if not, don't try to reverse?

cc @industrial-sloth

### DIFF
--- a/python/test/test_data.py
+++ b/python/test/test_data.py
@@ -255,6 +255,7 @@ class TestSeriesGetters(PySparkTestCase):
         # passing a range that is completely out of bounds throws a KeyError
         assert_raises(KeyError, self.series.__getitem__, (slice(2, 3), slice(None, None)))
 
+
 class TestDataMethods(PySparkTestCase):
 
     def test_sortbykey(self):
@@ -282,7 +283,7 @@ class TestDataMethods(PySparkTestCase):
         out = data.sortByKey().keys().collect()
         assert(array_equal(out, [(0,), (1,), (2,)]))
 
-    def collect(self):
+    def test_collect(self):
 
         dataLocal = [
             ((0, 0), array([0])),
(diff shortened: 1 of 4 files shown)


python/test/test_data.py
```diff
@@ -254,3 +254,72 @@ def test_brackets(self):
...
+            ((1, 2), array([5]))
+        ]
+
+        data = Data(self.sc.parallelize(dataLocal))
+
+        out = data.collectKeysAsArray()
+
+        assert(array_equal(out, [[0, 0], [0, 1], [0, 2], [1, 0], [1, 1], [1, 2]]))
+
+        out = data.collectKeysAsArray()
```
jwittenbach: I might be missing something, but I think this should be
```
out = data.collectValuesAsArray()
```
Not sure how it's passing the unit tests at the moment though....

jwittenbach:
From an API standpoint, I think this looks fantastic.

The implementation all seems to hang on the fact that, as long as you don't do any operation that causes a shuffle, the order of the returned values from `RDD.collect` will always be the same. Are we safe in assuming this?

python/test/test_data.py
jwittenbach: Speaking of which, I don't actually know how all of this works: does the Travis build automatically run the unit tests?

freeman-lab: Good catch, thanks! Yes, Travis automatically runs all the unit tests. I made a dumb mistake and didn't include the word `test` in this method, so this particular test wasn't running. Now that it's running, it's failing due to the line you flagged, but I just pushed a commit that corrects it.

freeman-lab:
Cleared this one point up offline with @jwittenbach , and tests are fixed and working, so will now merge in.

PR #113, opened 2015-02-18
Use tifffile.py with signed int tiffs

### LINKED ISSUE #110: "unrecognized mode" from multipage 16-bit signed tiff files
Attempting to load a 16-bit signed tiff file leads to a `ValueError: unrecognized mode`. This appears to be a problem with the underlying PIL / Pillow library (Pillow 2.7.0 currently on ec2) rather than with thunder itself:

```python
from PIL import Image
Image.VERSION  # returns "1.1.7"
Image.PILLOW_VERSION  # returns 2.7.0
im = Image.open("an216166_2013_07_17_run_01_sbv_01_main_002.tif")  # example 16 bit signed integer tiff file
im.mode  # returns "I;16S"
im.seek(1)  # throws ValueError
```

industrial-sloth:
This PR brings in Christoph Gohlke's `tifffile.py` module (http://www.lfd.uci.edu/~gohlke/, http://www.lfd.uci.edu/~gohlke/code/tifffile.py.html) and uses it instead of PIL/Pillow when a tiff image is detected as having signed integer values. This is expected to resolve #110. 

We may eventually want to consider using tifffile to replace PIL/Pillow more broadly, though currently this PR only addresses the specific failure mentioned in #110.


industrial-sloth:
It occurs to me I should probably check that images are coming out with the correct orientation; there may be an X/Y transpose that will need fixing.

broxtronix:
Hey @industrial-sloth... I just wanted to chime in to say that I'm highly in support of wrapping libtiff directly like this.  TIFF files come in such a variety of flavors (it's something of a meta-fileformat), and I have yet to find a third party image file reader whose API is flexible enough to expose all of the TIFF features.    We have been using pylibtiff (https://code.google.com/p/pylibtiff/), but Chris Gohlke's library looks just as good.     In any case, thumbs up to this patch!

industrial-sloth:
Thanks @broxtronix ! Yeah, PIL/Pillow's support for all the flavors of tiff file that one might find in scientific applications appears lacking - understandable, I suppose, since this really isn't their primary use case. It might be nice to do a quasi-formal, systematic shootout between all these different alternative tiff file readers, and see how well each covers the range of files that we're encountering. (I'm not 100% sure that even libtiff covers all the dialects our users are bringing in.) At the moment, this is definitely something of a hacky patch over one known broken case in PIL.

industrial-sloth:
I ran into issue #116 in testing this patch; even though that's really a distinct issue, I'm leaning towards folding the fix for that into this PR as well. The problem is that the data set in which I ran into #116 is made up of signed int tiffs, so I need this fix in order to test on that data. I suppose I could do a separate PR, branching off of this branch, but that seems like it would end up being messier than just combining two fixes in this single PR. 

If there are strong objections to this approach, please let me know soon. :)

python/thunder/rdds/fileio/imagesloader.py
```diff
@@ -199,35 +199,56 @@ def fromTif(self, dataPath, ext='tif', startIdx=None, stopIdx=None, recursive=Fa
...
+                if nplanes is not None:
+                    values = [ary[i:(i+nplanes)] for i in xrange(0, ary.shape[0], nplanes)]
+                else:
+                    values = [ary]
+                tfh.close()
+                # transpose Z dimension if any, leave X and Y in same order
+                if ary.ndim == 3:
+                    values = [val.transpose((1, 2, 0)) for val in values]
+                    # squeeze out last dimension if singleton
+                    values = [val.squeeze(-1) if val.shape[-1] == 1 else val for val in values]
```
freeman-lab: Are we doing this in the other case as well?

freeman-lab:
@industrial-sloth left one question, otherwise LGTM! Thanks! And I agree with both you and @broxtronix that wrapping is good, and at some point we need a full head-to-head on all this different options. But with this patch we should have a lot of our bases covered (famous last words...)

python/thunder/rdds/fileio/imagesloader.py
industrial-sloth: Offline I confirmed with @freeman-lab that "this" in this case is "squeezing out the last dimension if it's a singleton", and yes, that's the behavior that we expect. It's asserted (among possibly other cases) in `test_imagesloader:TestImagesFileLoaders.test_fromMultipleMultiTimepointTifs`, where we assert that the expected shape of a multipage tif loaded with `npages=1` is `(70, 75)`. So I think we're good here. :)

industrial-sloth:
Sweet, thanks @freeman-lab !

Since I don't yet have the fix for #116 ready to go, I'll go ahead and merge in this PR right now, and will plan to put that fix in another PR. But since this is getting merged into master already, the merge for that upcoming PR should be pretty smooth. My concerns in the comment above about wanting to get two fixes in this single PR are thus totally addressed. :)

PR #114, opened 2015-02-18
Add programmatic setting of AWS credentials

industrial-sloth:
AWS credentials are required to access S3. Typically these credentials end up in two locations - first, they're put in a `core-site.xml` file, where they're read out and used by the Hadoop InputFormat readers, which in Thunder primarily come in when reading binary Series data. For other access to S3, we use the `boto` module, which looks for credentials in a couple different places, including environment variables and a `.boto` config file. In the `thunder-ec2` script we write out these credentials to a `.boto` file and push this file out to all workers.

So far so good. The problem comes in when running in managed environments that don't give direct access to the cluster node file systems, such as the new Databricks Cloud. Here we can push out credentials to be used by the Hadoop readers by something like the following:
```python
sc._jsc.hadoopConfiguration().set("fs.s3n.awsAccessKeyId", ACCESS_KEY)
sc._jsc.hadoopConfiguration().set("fs.s3n.awsSecretAccessKey", SECRET_KEY)
```
But we don't have a good solution (or any solution, really) for getting the AWS credentials out where they can be used by boto on the workers. So, at the moment, on Databricks Cloud most attempts to access S3 fail once a worker node attempts to connect and discovers that it doesn't have the AWS credentials.

The "solution" implemented in this PR is to add a new `setAWSCredentials` method on the ThunderContext object (accessed as `tsc` in the shell). Calling this method should make it so that all data loaded in the future will use the passed access keys to reach S3. These keys should be propagated out to child Data objects, and in general everything should Just Work. If this method is *not* called, then the current behavior should be maintained - the only reason one should have to use this new method is if you are in an environment where you don't have direct access to set configuration files or environment variables on worker nodes.

### DIFF
--- a/python/test/test_seriesloader.py
+++ b/python/test/test_seriesloader.py
@@ -323,7 +323,7 @@ class TestSeriesBinaryLoader(PySparkTestCaseWithOutputDir):
 class TestSeriesBinaryWriteFromStack(PySparkTestCaseWithOutputDir):
 
     def _run_roundtrip_tst(self, testCount, arrays, blockSize):
-        print "Running TestSeriesBinaryWriteFromStack roundtrip test #%d" % testCount
+        #  print "Running TestSeriesBinaryWriteFromStack roundtrip test #%d" % testCount
         inSubdir = os.path.join(self.outputdir, 'input%d' % testCount)
         os.mkdir(inSubdir)
 
--- a/python/thunder/factorization/nmf.py
+++ b/python/thunder/factorization/nmf.py
@@ -202,6 +202,7 @@ class NMF(object):
 
             # report results
             self.h = h
+            # TODO: need to propagate metadata through to this new Series object
             self.w = Series(w)
 
         else:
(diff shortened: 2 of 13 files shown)


industrial-sloth:
This PR has a whole lot of plumbing, but not much in the way of actual meaningful changes. I'm planning to do some more extensive testing on ec2 tonight and tomorrow, but wanted to get this PR out now in its current state for visibility.

industrial-sloth:
After some testing on ec2 with all credentials saved in files blown away, I'm reasonably confident that this PR is complete and functional.

freeman-lab:
After a chat with @industrial-sloth offline, to handle the saving of data objects back to S3, we're going to try using the credentials as stored on the Hadoop configuration attached to the data object's rdd, as opposed to storing them explicitly as attributes on the data object.

industrial-sloth:
I'm going to close out this PR and open a new one after squashing commits.

Issue #116, opened 2015-02-19
varying numbers of images per file lead to incorrect Images keys


industrial-sloth:
We now support (in master) single image files that contain multiple logical images via the `nplanes` argument. The keys of the resulting `Images` objects are calculated assuming that each file will contain the same number of logical indexes - lines 254 and 255 in the version of ImagesLoader that I have in my `tmp` branch:
```python
nvals = len(values)
keys = [idx*nvals + timepoint for timepoint in xrange(nvals)]
```
This implicitly assumes that `nvals` will be the same for all image files. `Idx` is the zero-based count of the current image file, which allows `idx*nvals` to accurately determine the starting offset of the current file within the overall set of logical images, so long as `nvals` is fixed and identical for every file.

When this assumption is violated, as for instance if a single file contains all images with a variable-length trial, then the keys for the resulting Images object get all screwed up. This leads to problems downstream when converting to Series, because the Images keys are assumed to accurately reflect position of the image record within the overall RDD.

Fixing this I think will require another small set of jobs to calculate the total number of images and the number of images per file. This will only have to be run when `nplanes` is passed, because otherwise we know that a single file has a single image.

PR #117, opened 2015-02-19
Handle variable-size tifs

industrial-sloth:
This addresses #116 by explicitly counting the number of images loaded per file when opening tif images with `npages` specified. Previously we assumed that all files had the same number of images, which is both a restrictive assumption and impossible to enforce, since files are handled independently across the workers.

We don't need to recalculate `Images` keys when `npages` is left at its default value of `None`, since then we know we have one image per file. And currently we don't need to recalculate the keys when loading binary stack images, since there we specify a `dims` parameter up front and enforce that it applies for all files loaded. In the future we might want to consider lifting this restriction, since so long as we know an x and y dimension we should be able to infer z from the file size. If we were to make this change, at that point we'd then want to recalculate the Images keys in a similar way for stacks as well.

### DIFF
--- a/python/test/test_context.py
+++ b/python/test/test_context.py
@@ -246,3 +246,70 @@ class TestContextLoading(PySparkTestCaseWithOutputDir):
         assert_equals((6, 70, 75), rangeSeriesAry.shape)
         for idx in xrange(6):
             assert_true(array_equal(testimg_arys[idx % 3], rangeSeriesAry[idx]))
+
+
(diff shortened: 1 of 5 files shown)


industrial-sloth:
Or, we could do this by using `zipWithIndex`, because apparently that's a thing.

industrial-sloth:
Yeah, this way is easier. And better.

freeman-lab:
Perfect! LGTM, merging it in.

PR #118, opened 2015-02-20
S3 reads with recursive=True

industrial-sloth:
This PR enables the `recursive=True` flag when reading from S3. Previously this would have thrown a NotImplementedError.

After this change, loads like the following (which hits multiple nested directories in S3) appear to be working as expected:
```python
imgs = tsc.loadImages("s3n:///neuro.datasets/svoboda.lab/tactile.navigation/1/images-binary/", inputFormat="stack", ext="bin", recursive=True, stopIdx=654, dims=(512, 512, 4))
```

### DIFF
--- a/python/thunder/rdds/fileio/readers.py
+++ b/python/thunder/rdds/fileio/readers.py
@@ -297,13 +297,10 @@ class BotoS3ParallelReader(_BotoS3Client):
         return ["s3n:///%s/%s" % (bucketname, keyname) for keyname in keyNames]
 
     def _listFilesImpl(self, dataPath, ext=None, startIdx=None, stopIdx=None, recursive=False):
-        if recursive:
-            raise NotImplementedError("Recursive traversal of directories isn't yet implemented for S3 - sorry!")
         parse = _BotoS3Client.parseS3Query(dataPath)
-
         conn = boto.connect_s3()
         bucket = conn.get_bucket(parse[0])
-        keys = _BotoS3Client.retrieveKeys(bucket, parse[1], prefix=parse[2], postfix=parse[3])
+        keys = _BotoS3Client.retrieveKeys(bucket, parse[1], prefix=parse[2], postfix=parse[3], recursive=recursive)
         keyNameList = [key.name for key in keys]
         if ext:
             keyNameList = [keyname for keyname in keyNameList if keyname.endswith(ext)]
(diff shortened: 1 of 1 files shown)


python/thunder/rdds/fileio/readers.py
```diff
@@ -438,23 +432,16 @@ def __getMatchingKeys(self, dataPath, filename=None):
...
-    def list(self, dataPath, filename=None, startIdx=None, stopIdx=None, recursive=False, includeDirectories=True):
+    def list(self, dataPath, filename=None, startIdx=None, stopIdx=None, recursive=False, includeDirectories=False):
         """List s3 objects specified by dataPath.
 
         Returns sorted list of 's3n://' URIs.
         """
-        # TODO: note that the default value for includeDirectories is here True, which reflects the current (and
-        # longstanding) behavior of this class. This *differs* from the local FS list() method, which by default
-        # filters out directories (which is definitely what we want if we're reading with recursive=True).
-        # These two need to be made more consistent.
```
industrial-sloth: This TODO comment appears to have been inaccurate. This `list` function seems to have been filtering directory listings out of the returned results, regardless of the `includeDirectories` setting. I changed the default value here to False to better reflect the actual behavior, no matter what this comment claims, and made it so that if `includeDirectories=True` then directories will be returned (provided that there are actual S3 objects corresponding to the directory, and not just an implicit directory by the convention of using '/' as a path delimiter on S3).

python/thunder/rdds/fileio/readers.py
```diff
@@ -438,23 +432,16 @@ def __getMatchingKeys(self, dataPath, filename=None):
...
-        if recursive:
-            raise NotImplementedError("Recursive traversal of directories isn't yet implemented for S3 - sorry!")
-        if not includeDirectories:
-            raise NotImplementedError("Filtering out directories is not yet implemented for S3 - sorry!")
-
-        keys = self.__getMatchingKeys(dataPath, filename=filename)
+        keys = self.__getMatchingKeys(dataPath, filename=filename, excludeDirectories=not includeDirectories,
```
freeman-lab: Having a little trouble following the double negative / use of include vs exclude, it didn't make sense to just use one or the other throughout?

freeman-lab:
One minor comment, otherwise LGTM, just need to resolve the conflict (probably due to the AWS stuff I just merged).

freeman-lab:
Great, thanks! @industrial-sloth

PR #122, opened 2015-02-27
Casting scalars

### LINKED ISSUE #121: Error calling astype on singletons
Calling astype on a `Data` object with singleton entries yields an error. To reproduce, use:

```
data = tsc.loadExample('fish-series')
data.seriesMean().astype('int').first()
```
The error is `TypeError: astype() takes no keyword arguments`. It appears that when calling `astype` on a `numpy.float64` (or other scalar), it doesn't take the same any additional arguments that work when calling it on an array. For example,

```
x = np.array([1.0,2.0,3.0])
x.astype('int', casting='unsafe')
```
works fine but this doesn't
```
x = np.float64(1.0)
x.astype('int', casting='unsafe')
```
One potential solution would be to check the length of the index before calling astype, and to only include the additional arguments if it's greater than 1.

industrial-sloth:
This PR adds support for numpy and python scalars as values of Data records in `Data.astype()`. Fixes #121.

### DIFF
--- a/python/thunder/rdds/data.py
+++ b/python/thunder/rdds/data.py
@@ -313,7 +313,7 @@ class Data(object):
         """
         if dtype is None or dtype == '':
             return self
-        from numpy import can_cast, ndarray
+        from numpy import ndarray
         from numpy import dtype as dtypeFunc
         if dtype == 'smallfloat':
             # get the smallest floating point type that can be safely cast to from our current type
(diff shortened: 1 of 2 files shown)


python/thunder/rdds/data.py
```diff
@@ -313,12 +313,29 @@ def astype(self, dtype, casting='safe'):
...
+            if isinstance(v, ndarray):
+                return v.astype(dtype_, casting=casting_, copy=False)
+            elif can_cast(v, dtype_, casting=casting_):
+                if hasattr(v, 'astype'):
+                    # numpy scalars have an 'astype' method, but don't have the "casting" kwarg:
+                    return v.astype(dtype_)
+                else:
+                    # turn ourself into a numpy scalar of the appropriate type
+                    # despite the name, 'asarray' will return a scalar if a scalar is passed
+                    return asarray(v, dtype=dtype_)
```
freeman-lab: Discussed offline with @industrial-sloth , we're going to simplify this by dropping the additional `hasattr` type, and just call `return asarray([v], dtype=dtype_)[0]`, which will convert to the appropriate type and return a valid singleton `numpy` object.

freeman-lab:
Thanks @industrial-sloth , LGTM!

Issue #108, opened 2015-02-15
Support for conf file loading binary images


freeman-lab:
The binary image format require one specification (`dims`), and frequently requires specifying the data type (`dtype`). We should make it possible to load these from a `conf.json` folder in the same folder as the `.bin` files, just as we do for Series data. The format can be very simple:

```
{
     "dims": [X, Y, Z],
     "dtype": "int16"
}
```

freeman-lab:
This is essentially solved by https://github.com/thunder-project/thunder/commit/aff1198ac45eb259d012d364782e3935ed6f73c9, closing issue

PR #123, opened 2015-02-27
support JSON configuration file in Images loading

industrial-sloth:
This PR adds support for an optional JSON configuration file to specify parameters when loading Images data, resolving #108.

This is roughly analogous to the existing support for a `conf.json` file when loading Series binary files, which has been in place for a while. I haven't made any effort in this PR to merge the two, but that would be I think a relatively simple and useful refactoring.

The parameters that are supported inside of conf.json are those that refer to the images and image files themselves (e.g. `dims`, `dtype`, `ext`, `recursive`, `nplanes`) and not parameters that refer to how they are to be converted to a Series (not `blockSize`, `renumber`, etc). This distinction is pretty clear in my mind, but I admit it might not be so obvious to a casual user.

### DIFF
--- a/python/test/test_context.py
+++ b/python/test/test_context.py
@@ -1,3 +1,4 @@
+import json
 import os
 import unittest
 import json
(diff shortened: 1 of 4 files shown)


freeman-lab:
@industrial-sloth finally revisiting this! I actually just pushed a commit (https://github.com/thunder-project/thunder/commit/aff1198ac45eb259d012d364782e3935ed6f73c9) that achieves related functionality but restricted only to specification of dimensions and numerical type, as these are the properties with which binary files in particular cannot be loaded (and are fundamentally under-specified without, so it makes sense to keep with the data). 

For simplicity, I think I'd rather wait on considering the full configuration specification support offered in this PR, so we can close this for now but maybe revisit later on.

PR #134, opened 2015-03-14
Script to upgrade python

npyoung:
I did this as two scripts: one does the upgrade, the other distributes the upgrade script and runs it on the slaves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ install:
 script:
     - export SPARK_HOME=`pwd`/spark-1.1.0-bin-hadoop1
     - cd python/test
-    - ./run_tests.sh
+    - ./run_tests.sh -v --verbosity=2
--- a/python/bin/upgrade-python
+++ b/python/bin/upgrade-python
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+cd /root
+
+echo "-----------------------------------------------------------------"
+echo "Upgrading python locally"
+echo "-----------------------------------------------------------------"
+thunder/python/bin/upgrade-python-local.sh
+
+echo "-----------------------------------------------------------------"
+echo "Upgrading python on all slave nodes"
+echo "-----------------------------------------------------------------"
+yum install pssh -y
+
+spark-ec2/copy-dir thunder/python/bin/upgrade-python-local.sh
+
+pssh -h $SPARK_HOME/conf/slaves -t 900 /root/thunder/python/bin/upgrade-python-local.sh
--- a/python/bin/upgrade-python-local.sh
+++ b/python/bin/upgrade-python-local.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# install build tools 
+yum install make automake gcc gcc-c++ kernel-devel git-core -y 
+ 
+# install python 2.7 and change default python symlink 
+yum install python27-devel -y 
+rm /usr/bin/python
+ln -s /usr/bin/python2.7 /usr/bin/python 
+ 
+# point yum to the right version of python
+sed -i s/python/python2.6/g /usr/bin/yum 
+sed -i s/python2.6/python2.6/g /usr/bin/yum 
+ 
+# install pip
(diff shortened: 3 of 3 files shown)


freeman-lab:
Thanks for putting this together @npyoung ! In stepping through the main updater script (on the driver) I hit the following error at `pip install numpy scipy six Pillow cython`. I was starting from an unmodified installation after running `thunder-ec2`, using master branch, and Spark 1.3. Have you encountered this? Maybe it can be safely ignored?

```
ImportError: No module named numpy.distutils.core

----------------------------------------
Command "/usr/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-
zp_2i1/scipy/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read()
.replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-v2l6IG-record/
install-record.txt --single-version-externally-managed --compile" failed with error code 
1 in /tmp/pip-build-zp_2i1/scipy
```

I was able to proceed with the rest of the installation without further trouble, and successfully started ipython (on the driver) and checked a few simple imports. Have not yet tested Thunder functionality.

broxtronix:
Hey guys,

I've been experimenting with installing Anaconda Python on my clusters and using that for Spark.  It seems to be working very well, and I think it might be a simpler way to get python2.7 running on ec2 clusters.  Here are the commands needed to get it working:

wget http://09c8d0b2229f813c1b93-c95ac804525aac4b6dba79b00b39d1d3.r79.cf1.rackcdn.com/Anaconda-2.1.0-Linux-x86_64.sh
rm -rf $HOME/anaconda && bash Anaconda-2.1.0-Linux-x86_64.sh -b && rm Anaconda-2.1.0-Linux-x86_64.sh
echo \'export PATH=\$HOME/anaconda/bin:\$PATH:\$HOME/spark/bin\' >> $HOME/.bashrc

Then you just need to specify which Python Spark should use by adding this to the spark-env.sh file on your cluster:

export PYSPARK_DRIVER_PYTHON=$HOME/anaconda/bin/ipython
export PYSPARK_PYTHON=$HOME/anaconda/bin/python

I've been very happy with this solution!

freeman-lab:
@broxtronix thanks this is super appealing! It would also match what we're now using for our testing environment, which would be terrific for testing and consistency. But these commands only update on the driver, right? Then you just do the same operation (minus setting the env variables) on all the workers using pssh? And approximately how long does it all take?

broxtronix:
Ah, right!  I left out the bit about sync'ing over to the slaves.  The fastest way to do that is to run 
```
copy-dir $HOME/anaconda
```
Syncing everything over takes just a sec.

The whole process from start to finish adds somewhere between 30 secs to a minute to the cluster setup process, I'd estimate, and you could probably win back most if not all of that time since this would eliminate the python-2.6 centric setup tasks that are currently carried out in `ec2.py`.  

One other thing... if you want to update any packages once anaconda in installed, you can use `conda`.  Here is the formula I've found works well with Thunder in this setup:
```
conda update --yes numpy scipy ipython
conda install --yes jsonschema pillow seaborn scikit-learn
```

This yields the latest and greatest numpy, scipy, and ipython (which are all static compiled so installing with `conda` is super fast), as well as Thunder's usual dependencies.

broxtronix:
And come to think of it, I think that the ENV variable may need to be set on the slaves as well, so it's probably a good idea to `copy-dir` the `spark-env.sh` file over after adding those python configuration bits.

freeman-lab:
Great, thanks for the info! That sounds very promising indeed, both in terms of the environment set up itself as well as the speed. If everyone agrees maybe we should close this PR and open a new one with some version of @broxtronix 's scripts to install Anaconda? Would be awesome if you want to put that together @broxtronix but I'm also happy to give it a stab.

freeman-lab:
After meeting with @broxtronix offline to work through the Anaconda solution, I'm going to close this PR in favor of another one coming soon with an Anaconda installation and setup. Thanks for the effort @npyoung ! I hope the new version should satisfy the same needs.

PR #139, opened 2015-03-23
Source extraction

freeman-lab:
This PR introduces the "source extraction" API for Thunder. Source extraction describes the problem of identifying spatially-structured and temporally-coherent sources in spatio-temopral data (for example, neurons in fluorescence imaging data).

An [iPython notebook](http://nbviewer.ipython.org/github/thunder-project/thunder-example-notebooks/blob/master/source-extraction-demo.ipynb) introduces the proposed API and walks through working toy examples.

The key design goals were:
- Providing a general, modular framework or "harness" that supports plugging in a variety of existing algorithmic approaches while exposing a simple, high-level interface
- Maintaining scalability and performance by distributing core operations where possible using Spark

The design is based on extensive discussions with @pkaifosh, @poolio, @broxtronix  , @logang, @speron, @sofroniewn, feedback welcome from all of them (and from others! ccing @jakirkham @tomsains )

Although the API is new, it uses lots of core functionality we have been building into Thunder (for example, the ability to express `Images` as distributed `Blocks`, the ability to extract signals from arbitrary spatial regions in either `Images` or `Series`, and new support for JSON serialization).

Summary of key additions:
- Source and SourceModel classes for creating, manipulating, saving/loading, and rendering sources
- Block based source extraction methods, including algorithms and mergers
- Feature based source extraction methods, including algorithms and creators
- Two very simple example implementations (a feature method, `localmax`, and a block method, `nmf`).

New dependencies:
- scikit-image (image morphology processing)
- zope.cachedescriptors (lazy property handling)

Outstanding issues:
- The trick used to expose keyword arguments at the component level (e.g. for algorithms and mergers) to the top-level factory method does not trigger ordinary keyword argument validation. Open to alternatives, so long as we can still expose a top-level factory method, which has great benefit for ease-of-use.
- The `BasicBlockMerger` is extremely minimal and should be enhanced.
- More algorithm-specific unit tests.
- More geometric operators and comparators for sources (expand, dilate, overlap).
- At least one or two more example methods (let's add one that imports an algorithm from SIMA?)


pkaifosh:
This is looking nice! One suggestion that I have is to support sequential application of multiple cleaners. I expect that people will commonly wish to combine a few different Cleaners. For example, someone might wish first to merge overlapping ROIs, then to remove ROIs that are below a certain size threshold, then remove ROIs that are not sufficiently round in shape. Someone else might wish to do some, but not all of these operations, as well as some other cleaning operations, such as smoothing the boundaries of the ROIs. This could be done either by allowing a SourceExtractionMethod to have multiple cleaners (e.g. in SIMA, we append [post-processing steps](http://www.losonczylab.org/sima/1.1/api/segment.html#post-processing-steps)), or by designing natural way to compose multiple cleaners into one (e.g. a CompositeCleaner subclass of Cleaner). Making this as natural as possible should encourage the development nice modular Cleaner subclasses that can be usefully combined.

freeman-lab:
Thanks @pkaifosh ! I think a CompositeCleaner or something like it is a great idea! If it's literally just going to be a list of Cleaners, I'd probably skip the subclass and just allow `SourceExtractionMethod` to accept a single Cleaner, or a list of Cleaners, as you say. One question: do you think these are typically commutative and associative operations? And if not, is there any checking we should do to ensure they are performed in the correct sequence?

pkaifosh:
I would not assume these operations to be commutative. For example, filtering ROIs based on a minimum size would have a different effect if done before instead of after merging overlapping ROIs. I would recommend that the operations simply be performed in the order specified by the user's construction of the list of Cleaners. In order that the user specifies should be valid, since the clean method's input and return type are the same, so I don't think there is anything that needs to be checked here.

freeman-lab:
Great, makes sense, sounds good. So technically a user could request a sequence of Cleaners that, in the order provided, have an unintended result, but I agree, because of the inputs/outputs it will always at least be a "valid" sequence, so it's up to the user to make sure it does something reasonable.

freeman-lab:
Thanks again for the review, @pkaifosh I added support for sequence of `Cleaners`, and more generally refactored the cleaning to be a little more generic (as it applies to all of these algorithms) and flexible (as you suggested). Along with several other enhancements and fixes.

I am going to merge this version in, but will create a few issues with simple extensions and enhancements that we should explore next.

PR #148, opened 2015-03-31
added new regression tutorial

jwittenbach:
used mouse example data to show regression/correlation analysis plus visualization of results using Colorize.


freeman-lab:
Thanks for adding this @jwittenbach !

PR #153, opened 2015-04-04
updated documentation in data.py, svd.py, ica.py, and pca.py and fixed some typos

d-v-b:
This PR attempts to clarify the documentation in the factorization classes `thunder.factorization.ica`, `thunder.factorization.pca`, and `thunder.factorization.svd` by adding text in the docstrings for these functions that detail the different options for the kwarg `svdMethod`. 

This PR also attempts to clarify the documentation for statistical methods on the `data` class, e.g., `data.mean()`, by indicating in their docstrings that these methods are computed over keys and that output is returned as an ndarray.

I also fixed a few small typos.

### DIFF
--- a/python/thunder/factorization/ica.py
+++ b/python/thunder/factorization/ica.py
@@ -22,8 +22,10 @@ class ICA(object):
         Number of independent components to estimate
 
     svdMethod : string, optional, default = "auto"
-        Which SVD method to use. If set to 'auto',
-        will select preferred method based on dimensionality.
+        Which SVD method to use.
+        If set to 'direct', will compute the SVD directly.
+        If set to 'em', will approximate the SVD using iterative expectation-maximization algorithm.
+        If set to 'auto', will select preferred method based on dimensionality of input data.
 
     maxIter : Int, optional, default = 10
         Maximum number of iterations
@@ -42,6 +44,10 @@ class ICA(object):
     `sigs` : RowMatrix, nrows, each array of shape (c,)
         Estimated independent components
 
+    See also
+    --------
+    SVD : singular value decomposition
+    PCA: principal components analysis
     """
 
     def __init__(self, c, k=None, svdMethod='auto', maxIter=10, tol=0.000001, seed=0):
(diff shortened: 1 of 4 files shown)


python/thunder/factorization/ica.py
```diff
@@ -22,8 +22,10 @@ class ICA(object):
         Number of independent components to estimate
 
     svdMethod : string, optional, default = "auto"
-        Which SVD method to use. If set to 'auto',
-        will select preferred method based on dimensionality.
+        Which SVD method to use.
+        If set to 'direct', will compute the SVD directly.
```
freeman-lab: replace "directly" with "direct gramian matrix estimation and eigenvector decomposition" here and in the other one

python/thunder/factorization/ica.py
```diff
@@ -22,8 +22,10 @@ class ICA(object):
...
+        If set to 'em', will approximate the SVD using iterative expectation-maximization algorithm.
+        If set to 'auto', will select preferred method based on dimensionality of input data.
```
freeman-lab: maybe add that auto means "use 'em' if number of columns is greater than 750"

python/thunder/rdds/data.py
```diff
@@ -459,7 +459,7 @@ def count(self):
 
     def mean(self, dtype='float64', casting='safe'):
         """
-        Mean of values, ignoring keys
+        Mean of values across keys, returned as an ndarray.
```
freeman-lab: These are better but might be even more clear as "Mean of values computed by aggregating across records, returned as an ndarray". A little funny to say mean across keys, almost implies that we're adding up the keys. You could further add that the result will be the same size as a single record. Seem good? Change the others accordingly?

freeman-lab:
Thanks @d-v-b ! Just left a couple minor suggestions, if those look good, once you change them will merge in.

d-v-b:
integrated your suggestions, let me know what you think!

freeman-lab:
Awesome, LGTM. Thanks @d-v-b !

Issue #156, opened 2015-04-06
Feature request: option to have selectByIndex return mask used in selection


jwittenbach:
`Series.selectByIndex` returns a new `Series` with only a subset of each record retained. In some cases, the user may have auxiliary series that have the same structure as those in the `Series` (e.g. covariates) and would like to apply the same selection to these stand-alone series as well. A simple way to achieve this is to optionally return the mask that `Series.selectByIndex` uses when doing the selection. This mask can then easily applied to series that are not included as records in the main `Series`.

PR #157, opened 2015-04-06
Select by index mask

jwittenbach:
Implements feature request #156.

Added an optional `returnMask` keyword argument to `Series.selectByIndex` which returns the mask used during the selection as an additional return value. Default value is `False`.


freeman-lab:
LGTM, thanks for the fix @jwittenbach !

jwittenbach:
I'm looking at the diff on this and something strange happened here. These were actually the changes that I had meant to submit in a different PR addressing the fact that my multi-index code was directly creating the resulting `Series` objects rather than using `Series._constructor`. The change referenced in this PR still needs to be made.

freeman-lab:
Oh oops, I somehow missed that. Feel free to submit a new PR, and probably change the title/description on this one.

jwittenbach:
Yeah, working on it now. Just going to add a unit test for the new functionality, and then I'll submit it.

PR #158, opened 2015-04-06
Google Cloud Storage support via Boto

broxtronix:
I have recently begun playing with Google Compute Engine, and I quite like it!   This is the first of hopefully several pull requests that will add support to Thunder for using Google Cloud services.

This merge request adds the capability to read data directly from Google Cloud Storage, which is the Google equivalent of Amazon S3.    Like S3, Google Storage can be accessed using the Boto python library.  This merge request mostly re-works the boto code in Thunder to be able to correctly handle both S3 and Google Storage endpoints.  The generalization was straight-forward, but I did go through the codebase and generally replaced class and variable names that contained 'S3' with 'Boto' since these classes now generalize to multiple URL endpoints.   With these changes, it is now possible to specify a URL of the form `gs://<bucket>/<key>` just like `s3://<bucket>/<key>`.  All previous functionality that was available when accessing S3 using Boto should now be available using Google Storage.

A note on Authentication:

Google Storage can authenticate a little differently from S3.  It is possible to specify an `ACCESS_KEY` and `SECRET_KEY` key for google storage in a `~/.boto` file, but on Google Compute Engine instances this is not necessary if the compute instance itself has already been authorized to access your storage resources.  This makes authentication very painless when accessing Google Storage from Google instances.  When access Google storage from elsewhere, you must explicitly supply a key pair.  We may want to discuss the typical use cases where users may want to authenticate with Google Storage, but my feeling is that this addresses the most important cases for now.

A note on HDFS:

This pull request does not (currently) enable access to Google Storage through the Hadoop filesystem.  This is possible, but setting this up requires installation of the [Google Storage adaptor for Hadoop](https://cloud.google.com/hadoop/google-cloud-storage-connector), and proper configuration of Hadoop itself.   This means that certain functions that rely on HDFS to access cloud storage like `loadSeries()` will not work out of the box with `gs://...` urls.   

This pull request has been tested on Google and Amazon, but could certainly benefit from a bit more testing and use before it gets merged in.  Comments and feedback are welcome!


broxtronix:
Also, I apologize for the two commits that are mostly reformatting tabs into spaces.  These were due to a bad configuration in my text editor that happened at an inconvenient time in the midst of making the very first commit. :)

On a related note, the Thunder style guide does not specify whether Thunder source files should use tabs or spaces for indentation.  I believe the predominant style in Python is to use four 4 spaces for indentation.  I think it would be worth adding a guideline to the style guide either way!

python/thunder/rdds/fileio/imagesloader.py
```diff
@@ -81,7 +81,7 @@ def fromStack(self, dataPath, dims=None, dtype=None, ext='stack', startIdx=None,
         recursive: boolean, default False
             If true, will recursively descend directories rooted at datapath, loading all files in the tree that
             have an extension matching 'ext'. Recursive loading is currently only implemented for local filesystems
-            (not s3).
+            (not s3 or Google Storage).
```
freeman-lab: Total nit, but maybe abbreviate `S3` and `GS` here to match the way you abbreviated the two elsewhere.

python/thunder/rdds/fileio/seriesloader.py
```diff
@@ -453,9 +453,11 @@ def readBlockFromTiff(planeIdxBlockIdx):
                 fp = reader_.open(fname)
                 try:
                     if doMinimizeReads:
-                        # use multitif module to generate a fake, in-memory one-page tif file
-                        # the advantage of this is that it cuts way down on the many small reads
-                        # that PIL/pillow will make otherwise, which would be a problem for s3
+                        # use multitif module to generate a fake, in-memory
+                        # one-page tif file the advantage of this is that it
```
freeman-lab: add a period between `one-page tif file` and `the advantage`

freeman-lab:
@broxtronix sorry this took awhile to get to, it's a fantastic patch! Very clean integration, and clearly paves the way towards Google Cloud integration with Thunder, which I'm very excited about.

I'm not yet fully setup on GC, but am doing some extra testing with this today on AWS. I trust your testing on the GC side, and especially if everything on AWS is unchanged (which it seems it should be), I'm happy merging this in. And I left a couple very minor style nits.

freeman-lab:
@broxtronix I'm going to merge this in, finally did some proper testing with this build on existing S3 workflows (e.g. loading one of the example codeneuro datasets on S3, via `params, data = tsc.loadExampleS3('svoboda.lab/tactile.navigation')` and everything worked fine. Thanks again, looking forward to further PRs on integrating Google Cloud + Thunder!

PR #160, opened 2015-04-08
Image-based local correlation

andrewosh:
The current version of localCorr, implemented in the SpatialSeries class, requires a significant amount of data duplication and shuffling, most of which can be optimized away. This version, suggested by @sofroniewn and @d-v-b, first does spatial averaging on an Images dataset, then appends the averaged data to the original before computing the correlation in time (by first converting to a Series and then correlating the first half of the series with the second). 

The localCorr unit tests have been ported from test_spatialseries.py to test_images.py, and all the relevant ones pass.

### DIFF
--- a/python/test/test_images.py
+++ b/python/test/test_images.py
@@ -650,7 +650,7 @@ class TestImagesLocalCorr(PySparkTestCase):
 
         corr = self.get_local_corr(dataLocal, 1)
 
-        assert(allclose(corr[4][1], truth))
+        assert(allclose(corr[1][1], truth))
 
     def test_localCorr_0Indexing_3D(self):
 
@@ -673,12 +673,9 @@ class TestImagesLocalCorr(PySparkTestCase):
 
         corr = self.get_local_corr(dataLocal, 1)
 
-        assert(allclose(corr[4][1], truth))
+        assert(allclose(corr[1][1], truth))
 
     def test_localCorr_Images_2D(self):
-        """
-        Test the localCorr without converting from Series to Images (using self.get_local_corr) first.
-        """
 
         dataLocal = [
             (0, array([[1.0, 2.0, 9.0], [5.0, 4.0, 5.0], [4.0, 6.0, 0.0]])),
@@ -694,14 +691,9 @@ class TestImagesLocalCorr(PySparkTestCase):
 
         corr = self.get_local_corr(dataLocal, 1, images=True)
 
-        assert(allclose(corr[4][1], truth))
+        assert(allclose(corr[1][1], truth))
 
     def test_localCorr_Images_3D(self):
-        """
-        Test the localCorr without converting from Series to Images (using self.get_local_corr) first.
-        """
-
-        #TODO
 
         dataLocal = [
             (0, array([[1.0, 2.0, 9.0], [5.0, 4.0, 5.0], [4.0, 6.0, 0.0]])),
@@ -717,7 +709,7 @@ class TestImagesLocalCorr(PySparkTestCase):
 
         corr = self.get_local_corr(dataLocal, 1, images=True)
 
-        assert(allclose(corr[4][1], truth))
+        assert(allclose(corr[1][1], truth))
 
 
 class TestImagesUsingOutputDir(PySparkTestCaseWithOutputDir):
--- a/python/test/test_images.py
+++ b/python/test/test_images.py
@@ -1,7 +1,7 @@
 import glob
 import struct
 import os
-from numpy import allclose, arange, array, array_equal, prod, squeeze, zeros
+from numpy import allclose, arange, array, array_equal, prod, squeeze, zeros, mean, corrcoef
 from numpy import dtype as dtypeFunc
 import itertools
 from nose.tools import assert_equals, assert_raises, assert_true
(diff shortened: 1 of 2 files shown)


python/test/test_images.py
```diff
@@ -613,6 +613,67 @@ def test_meanWithSingleRegionIndices3D(self):
         assert_equals(5, collected[0][1][0])
         assert_equals(15, collected[1][1][0])
 
+class TestImagesLocalCorr(PySparkTestCase):
```
freeman-lab: These tests are good, but the conversion from Series to Images is bit indirect. I'd like to add at least versions of these two tests that explicitly construct data with the same content as `Images` objects, and perform the same operation.

python/thunder/rdds/images.py
```diff
@@ -436,6 +436,41 @@ def medianFilter(self, size=2):
...
+
+        This algorithm computes, for every spatial record, the correlation coefficient
+        between that record's series, and the average series of all records within
+        a local neighborhood with a size defined by the neighborhood parameter.
+        For data with three spatial keys, only neighborhoods in x and y
+        currently supported.
+
+        parameters
+        ----------
+        neighborhood: int, optional, default=2
```
freeman-lab: Handling sequences will be tricky here. Currently, if you pass a sequence of two numbers to the filtering operation for 3D data, it will fail because it assumes one number per dimension. The same thing will happen here if neighborhood is a list of two numbers. Ultimately, we should modify the filtering methods to handle as a special case a 2-number sequence for 3D data, and then do the same here, but for now I would say just force the neighborhood to be a scalar.

python/thunder/rdds/images.py
```diff
@@ -436,6 +436,41 @@ def medianFilter(self, size=2):
...
+        blurred = self.uniformFilter((neighborhood * 2) + 1)
+
+        # Union the averaged images with the originals to create an Images object containing 2N images (where
+        # N is the original number of images), ordered such that the first N images are the averaged ones.
+        combined = self.rdd.union(blurred.rdd)
+        combinedImages = self._constructor(combined, nrecords=(2 * nimages)).__finalize__(self).renumber()
+
+        # Correlate the first N (averaged) records with the last N (original) records
+        series = combinedImages.toSeries()
+        corr = series.applyValues(lambda x: corrcoef(x[:nimages], x[nimages:])[0, 1])
```
freeman-lab: After thinking about it more, I think we might want to just do the pack here. Conceptually this kind of like calling `.mean()` or `.stdev()` on an `Images` object, because it collapses across images to yield a single number per pixel. There's also something slightly weird about returning a `Series` object from an `Images` operation.

At least, we could make `pack` or something a boolean argument and default to True.

freeman-lab: After discussing more offline with @andrewosh and @jwittenbach , we're going to pack and not make it an argument.

python/thunder/rdds/images.py
```diff
@@ -436,6 +436,41 @@ def medianFilter(self, size=2):
...
+
+        nimages = self.nrecords
+
+        # Spatially average the original image set over the specified neighborhood
+        blurred = self.uniformFilter((neighborhood * 2) + 1)
+
+        # Union the averaged images with the originals to create an Images object containing 2N images (where
+        # N is the original number of images), ordered such that the first N images are the averaged ones.
+        combined = self.rdd.union(blurred.rdd)
+        combinedImages = self._constructor(combined, nrecords=(2 * nimages)).__finalize__(self).renumber()
```
freeman-lab: Instead of calling `renumber` here, use the trick above so that these the correct keys already.

python/thunder/rdds/images.py
```diff
@@ -436,6 +436,41 @@ def medianFilter(self, size=2):
...
+            Size of the correlation neighborhood in pixels. A sequence is interpreted
+            as the neighborhood size for each axis. For three-dimensional data, a single
+            scalar is intrepreted as the neighborhood in x and y, with no filtering in z.
+        """
+        from numpy import corrcoef
+
+        nimages = self.nrecords
+
+        # Spatially average the original image set over the specified neighborhood
+        blurred = self.uniformFilter((neighborhood * 2) + 1)
```
freeman-lab: Add another step after this that uses `applyKeys` to increment the keys by nimages, that way the keys will already be correct in the next step and we don't need to use `renumber()`.

poolio:
Another possibility: you could work only on `Images` and avoid the conversion to `Series` by  computing the uncentered moments (see [here](http://en.wikipedia.org/wiki/Pearson_product-moment_correlation_coefficient)). Maybe something like:
```python
mu = self.mean()
muAvg = blurred.mean()
stdev = self.stdev()
stdevAvg = blurred.stdev()
muCross = Images(self.rdd.mapValues(lambda x: x * uniformFilter(x, ...))).mean()

localCorr = (muCross - mu * muAvg) / (stdev * stdevAvg)
````
You could also merge the moment computations all into one mapreduce call so you don't have to recompute the blurred image, but then you'll have a really massive intermediate dataset.

freeman-lab:
@poolio cool suggestion! I think ultimately we'll have to test both approaches on some large data sets. 

My hunch is that, at least for uncached data, the version that converts to `Series` might be faster because it only requires one pass over the data. We could definitely combine the aggregations with the existing `stats` method, but yes, the aggregations could then become unwieldy, especially for large volumes. And it'd still be three full passes. But especially for cached data, could definitely be a win.

One idea I had for the current PR based on your suggestion is that the conversion to `Series` could actually just be a conversion to `Blocks`, and the correlation would be computed on the blocks. That would prevent instantiating a record per pixel (the final step of the `Series` conversion), and more importantly, be a potentially faster correlation computation (by exploiting matrix math).

freeman-lab:
Changes are great, LGTM, thanks @andrewosh ! Merging it in.

Btw, @poolio , definitely still interested in possible performance optimizations. I'd like to merge this version in now, as it's already a significant improvement over the existing pure `Series` domain approach, but we should definitely continue to explore alternatives!

PR #166, opened 2015-04-12
Pull request for basic distance and overlap methods between sources

logang:
added pull request for changes to extraction/source.py that include basic distance metrics for sources

### DIFF
--- a/python/thunder/extraction/source.py
+++ b/python/thunder/extraction/source.py
@@ -1,7 +1,6 @@
 from numpy import asarray, mean, sqrt, ndarray, amin, amax, concatenate, sum, zeros, maximum, \
     argmin, newaxis, ones, delete, NaN, inf, isnan
 
-import numpy.linalg as la
 from scipy.stats import spearmanr
 
 from thunder.utils.serializable import Serializable
@@ -109,7 +108,7 @@ class Source(Serializable, object):
                 del self.__dict__[prop]
         return self
 
-    def distance(self, other, method='l1'):
+    def distance(self, other, method='euclidean'):
         """
         Distance between the center of this source and another.
 
@@ -121,17 +120,17 @@ class Source(Serializable, object):
             L1-norm ('l1'). 
 
         """
+        from numpy.linalg import norm
         if method == 'euclidean':
             order = 2
         elif method == 'l1':
             order = 1
         else:
             raise ValueError('Given distance method not supported.')
-
         if isinstance(other, Source):
-            return la.norm(self.center - other.center, ord=order)
+            return norm(self.center - other.center, ord=order)
         elif isinstance(other, list) or isinstance(other, ndarray):
-            return la.norm(self.center - asarray(other), ord=order)
+            return norm(self.center - asarray(other), ord=order)
 
     def overlap(self, other, method='support',counts=False):
         """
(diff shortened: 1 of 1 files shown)


python/thunder/extraction/source.py
```diff
@@ -1,6 +1,9 @@
 from numpy import asarray, mean, sqrt, ndarray, amin, amax, concatenate, sum, zeros, maximum, \
     argmin, newaxis, ones, delete, NaN, inf, isnan
 
+import numpy.linalg as la
+from scipy.stats import spearmanr
```
freeman-lab: This looks like it's unused

python/thunder/extraction/source.py
```diff
@@ -106,18 +109,66 @@ def restore(self, skip=None):
...
+            self.coordinates = self.coordinates.tolist()
+        if isinstance(other.coordinates, ndarray):
+            other.coordinates = other.coordinates.tolist()
+        intersection = [a for a in self.coordinates if a in other.coordinates]
+        complement = [a for a in self.coordinates if a not in intersection]
+        if method == 'support':
+            hits = len(intersection)
+            misses = len(complement)
+            if counts:
+                return (hits,misses)
```
freeman-lab: insert space before `misses`

python/thunder/extraction/source.py
```diff
@@ -1,6 +1,9 @@
 from numpy import asarray, mean, sqrt, ndarray, amin, amax, concatenate, sum, zeros, maximum, \
     argmin, newaxis, ones, delete, NaN, inf, isnan
 
+import numpy.linalg as la
```
freeman-lab: Can we move this one so it's locally scoped to `distance`, and use `from numpy.linalg import norm`?

python/thunder/extraction/source.py
```diff
@@ -106,18 +109,66 @@ def restore(self, skip=None):
...
+        """
+        Compute the overlap between current source and 'other' source, in terms 
+        of either support or similarity of coefficients. 
+
+        Parameters
+        ----------
+        method : str
+            Compare either support of source coefficients ('support'), or the 
+            source spatial filters (not yet implemented).
+
```
freeman-lab: Remove empty line

python/thunder/extraction/source.py
```diff
@@ -106,18 +109,66 @@ def restore(self, skip=None):
...
+            #TODO: this is actually a bit complicated, as support size can matter... we
+            # could use a fixed box size or the union of source supports?
+            raise NotImplementedError()
+        else:
+            raise ValueError('Given method is not supported.')
 
     def tolist(self):
         """
         Convert array-like attributes to list
+
```
freeman-lab: Remove empty line

python/thunder/extraction/source.py
```diff
@@ -106,18 +109,66 @@ def restore(self, skip=None):
...
+        intersection = [a for a in self.coordinates if a in other.coordinates]
+        complement = [a for a in self.coordinates if a not in intersection]
+        if method == 'support':
+            hits = len(intersection)
+            misses = len(complement)
+            if counts:
+                return (hits,misses)
+            else:
+                return hits/float(hits+misses)
+        elif method == 'corr':
```
freeman-lab: Or intersection? Specifically, take the intersection computed above, and compute a correlation coefficient on just those values?

freeman-lab:
Thanks @logang ! This is looking really good. I left some minor comments, a couple additional things:

- Can you figure out why the tests are failing? To run them locally, just follow these [instructions](http://thunder-project.org/thunder/docs/contributing.html#running-the-tests). Something's broken in the distance calculation, but I can't quite tell how. It might be changing the default from `euclidean` to `l1`, in which case, we'll need to change the tests, or change the default. I think I have a slight preference for restoring `euclidean` as the default.

- I'd like to add something for the `corr` method before merging, rather than leave as a todo.

- In your PR we've lost the `similarity` function, which was designed to estimate the fraction of sources in one model also found in another, where "found in" is defined as having a value less than some threshold for one of the comparison operations. Can we add that back in? Maybe you found the name confusing? Open to alternatives. It should probably look like the one from before, but now include both `distance` and `overlap` as possible arguments for `metric`, which will be great to have!

freeman-lab:
@logang just pinging you on this one =) would be awesome to merge in soon, as we've made a bunch of progress over on [neurofinder](https://github.com/CodeNeuro/neurofinder) and would love to use these metrics in the initial testing stages!

python/thunder/extraction/source.py
logang: ok so we will do: (\rho*hits)/(hits+misses) where \rho is the spearman correlation

freeman-lab:
Awesome, LGTM, thanks for the contribution @logang !

PR #168, opened 2015-04-13
BlockMethod to run any SIMA segmentation strategy

pkaifosh:
I've created a simple BlockMethod/BlockAlgorithm pair that should allow any sima segmentation strategy to be run on a thunder block. Included is a single test case, which only runs if sima is installed. The test case passes on my laptop.

On a related note, I may try to simplify the dependencies of SIMA a bit so that it might be made a dependency of thunder without much pain. The current dependencies are:

    Python 2.7
    numpy >= 1.6.2
    scipy >= 0.13.0
    scikit-image >= 0.9.3
    scikit-learn >= 0.11
    shapely >= 1.2.14
    pillow >= 2.6.1
    future

@freeman-lab Let me know which ones you think would have to best converted to optional dependencies prior to making sima a thunder dependency.

### DIFF
--- a/python/thunder/extraction/block/methods/sima.py
+++ b/python/thunder/extraction/block/methods/sima.py
@@ -1,0 +1,50 @@
+from thunder.extraction.block.base import BlockMethod, BlockAlgorithm
+from thunder.extraction.source import Source
+
+
+class BlockSIMA(BlockMethod):
+
+    def __init__(self, simaStrategy, **kwargs):
+        algorithm = SIMABlockAlgorithm(simaStrategy, **kwargs)
+        super(self.__class__, self).__init__(algorithm, **kwargs)
+
+
+class SIMABlockAlgorithm(BlockAlgorithm):
+    """
+    Find sources using a SIMA SegmentationStrategy.
+
+    Parameters
+    ----------
+    sima_strategy : sima.segment.SegmentationStrategy
+        The particular strategy from SIMA to be used.
+    """
+    def __init__(self, simaStrategy, **extra):
+        self.strategy = simaStrategy
+
+    def extract(self, block):
+        import sima
+        import numpy
+
+        # reshape the block to (t, z, y, x, c)
+        dims = block.shape
+        if len(dims) == 3:  # (t, x, y)
+            reshapedBlock = block.reshape(dims[0], 1, dims[2], dims[1], 1)
+        else:  # (t, x, y, z)
+            reshapedBlock = block.reshape(dims[0], dims[3], dims[2], dims[1], 1)
+
+        # create SIMA dataset from block
+        dataset = sima.ImagingDataset([sima.Sequence.create('ndarray', reshapedBlock)], None)
+
+        # apply the sima strategy to the dataset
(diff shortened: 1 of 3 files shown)


python/thunder/extraction/block/methods/sima.py
```diff
@@ -0,0 +1,50 @@
...
+        super(self.__class__, self).__init__(algorithm, **kwargs)
+
+
+class SIMABlockAlgorithm(BlockAlgorithm):
+    """
+    Find sources using a SIMA SegmentationStrategy.
+
+    Parameters
+    ----------
+    sima_strategy : sima.segment.SegmentationStrategy
```
freeman-lab: Change `sima_strategy` in docstring to match `simaStrategy` as below

python/thunder/extraction/block/methods/sima.py
```diff
@@ -0,0 +1,50 @@
...
+            reshapedBlock = block.reshape(dims[0], dims[3], dims[2], dims[1], 1)
+
+        # create SIMA dataset from block
+        dataset = sima.ImagingDataset([sima.Sequence.create('ndarray', reshapedBlock)], None)
+
+        # apply the sima strategy to the dataset
+        rois = self.strategy.segment(dataset)
+
+        # convert the coordinates between the SIMA and thunder conventions
+        coords = [numpy.asarray(numpy.where(numpy.array(roi))).T for roi in rois]
```
freeman-lab: our convention has been to explicitly import from numpy, as in `from numpy import asarray, where`, and probably just put at the very top (before the thunder imports)

freeman-lab:
This looks great @pkaifosh , thanks! Only left a couple minor formatting requests.

Regarding making it a dependency, it looks like it will be quite smooth now, thanks for the effort! All of the ones you list are also dependencies for thunder except `future` and `shapely`. Adding `future` seems fine, and I had been thinking about adding `shapely` anyway. So far every time I've considered it I found I could do similar things with `matplotlib` or `skimage`, though I imagine that won't always be true. Just out of curiosity, where have you found `shapely` to be crucial?

Why don't we plan to merge your PR as it is now (with SIMA as an optional dependency), then I can make it a full dependency once I've done some testing to make sure everything installs fine on our EC2 builds. Sounds good?

pkaifosh:
Thanks for the feedback, @freeman-lab . I'll make the changes that you have suggested, and then you can merge the PR.

Regarding shapely, we use it for the polygon representations of our ROIs. I doubt it is crucial, as we use only very limited functionality from it, and I really wouldn't mind removing the dependency since it has caused installation headaches for some of our Windows users due to DLL issues. We just haven't got around to substituting it out though because, other than the installation issues on Windows, it works as is, and there are so many other things to do.

freeman-lab:
Awesome, this LGMT @pkaifosh ! Thanks for the contribution! First merged PR from CodeNeuro =)

And maybe I can help with that shapely dependency, I think we're using a scipy function for getting polygons from masks (but looked at something similar in shapely), maybe that's all you need too, I can take a look.

PR #170, opened 2015-04-16
Implement sample data from factor analysis model.

rhofour:
This allows generating high dimensional example data which
comes from a lower dimensional subspace. Originally I intended
to implement factor analysis, but I think this new example data is
useful on its own.

### DIFF
--- a/python/thunder/utils/context.py
+++ b/python/thunder/utils/context.py
@@ -506,7 +506,7 @@ class ThunderContext():
         """
         Make an example data set for testing analyses.
 
-        Options include 'pca', 'fa', 'kmeans', and 'ica'.
+        Options include 'pca', 'factor', 'kmeans', and 'ica'.
         See thunder.utils.datasets for detailed options.
 
         Parameters
--- a/python/thunder/utils/context.py
+++ b/python/thunder/utils/context.py
@@ -506,7 +506,7 @@ class ThunderContext():
         """
         Make an example data set for testing analyses.
 
-        Options include 'pca', 'kmeans', and 'ica'.
+        Options include 'pca', 'fa', 'kmeans', and 'ica'.
         See thunder.utils.datasets for detailed options.
 
         Parameters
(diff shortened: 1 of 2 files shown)


python/thunder/utils/datasets.py
```diff
@@ -63,6 +63,44 @@ def generate(self, k=3, npartitions=10, nrows=100, ncols=10, seed=None):
         else:
             return data
 
+class FAData(DataSets):
```
freeman-lab: I find `FA` slightly ambiguous, and I'm not sure it's quite well known enough by it's abbreviation, compared to say PCA and ICA. I might change the class name to `FactorAnalysisData` and then use `factor` as the shorthand.

python/thunder/utils/datasets.py
```diff
@@ -63,6 +63,44 @@ def generate(self, k=3, npartitions=10, nrows=100, ncols=10, seed=None):
...
+
+    def generate(self, q=1, p=3, nrows=50, npartitions=10, sigmas=None, seed=None):
+        """
+        Generate data from a factor analysis model
+
+        Parameters
+        ----------
+        q : int
```
freeman-lab: Indicate the default, here and elsewhere, as in `q : int, optional, default = 1`

python/thunder/utils/datasets.py
```diff
@@ -137,6 +175,7 @@ def pointToCircle(center, radius):
 DATASET_MAKERS = {
     'kmeans': KMeansData,
     'pca': PCAData,
+    'fa': FAData,
```
freeman-lab: As suggested above, probably change to `'factor': FactorAnalysisData`

python/thunder/utils/context.py
```diff
@@ -506,7 +506,7 @@ def makeExample(self, dataset, **opts):
         """
         Make an example data set for testing analyses.
 
-        Options include 'pca', 'kmeans', and 'ica'.
+        Options include 'pca', 'fa', 'kmeans', and 'ica'.
```
freeman-lab: And also change to `factor` here

freeman-lab:
Thanks for this! Will be cool to have. I just made some suggestions about the naming conventions, and about the docstrings, otherwise looks good!

PR #174, opened 2015-04-20
Update faq.rst

vjlbym:


GrantRVD:
I think the `-` need to be included in the arguments for it to be a valid environment variable argument, so the lines should be `export _JAVA_OPTIONS='-Xms512m -Xmx1024m'` instead of `export _JAVA_OPTIONS='Xms512m Xmx1024m'`. I could be wrong but I don't think the latter setting would do anything functional because of syntax.

It might be helpful to break up this answer into solutions for UNIX and solutions for Windows. I can add a couple ways to set the environment variable on Windows after this pull request is merged, if no one minds.

vjlbym:
Yeah, you're right. Sorry, that was a typo. Good catch.

vjlbym:
Yeah, it'd be good to split it into Windows and Unix/MacOS. What I wrote should work for the latter two, I believe. Is the Windows way Control Panel->System->Advanced System Settings->Change Environment Variables->New ?

GrantRVD:
That's the manual point-and-click way to do it - we should probably stick to the command-line ways to keep it relatively the same.

If I'm not mistaken, the `set` command should work in the regular Windows cmd prompt, Just replace `export` with `set` and drop the quotes. In addition, some people might be using PowerShell, like me, which allows both the `set` command and setting environment variables using .NET. All in all, I think the breakdown for setting the initial heap space to 512 MB and max heap space to 1 GB as in our example would be
```PowerShell
// UNIX/MacOS
export _JAVA_OPTIONS='-Xms512m -Xmx2g'

// Windows (cmd or PowerShell)
set _JAVA_OPTIONs=-Xms512m -Xmx1g

// Windows PowerShell (.NET alternative)
[environment]::SetEnvironmentVariable("_JAVA_OPTIONS","-Xms512m -Xmx1g","Machine")
```
On windows terminals at least, changes to the Environment Variables may not display until the shell is restarted.

vjlbym:
Yeah, `set` should work based on my experience. You have to start a new terminal in Unix too. If you make a system wide environment variable, you actually have to log out for it to take effect.

python/doc/faq.rst
```diff
@@ -48,6 +48,10 @@ Basic usage
...
 
+*I'm getting "java.lang.OutOfMemoryError: Java heap space errors" during usage on a local machine. What should I do?*
+	
+	This error indicates that the JAVA heap space is being exceeded by Thunder. One solution is to set JAVA runtime options to use more heap space. To make this change permanent when calling JAVA, open your bash profile (e.g. ``~/.bashrc`` on Ubuntu) and add a line similar to ``export _JAVA_OPTIONS="-Xms512m -Xmx4g"``. In this example, the initial heap space is set to 512mb and the max heap space is set to 4gb. Note here to use ``_JAVA_OPTIONS`` and not ``JAVA_OPTS``. ``JAVA_OPTS`` is not an environment variable that JDK recognizes on its own, but is used by other apps while running JAVA.
```
freeman-lab: Based on the discussion here, I'd rewrite the sentence starting with "To make this change permanent..." as "To make this change permanent when calling any applications using JAVA, open your bash profile (e.g. ``~/.bash_profile`` on Mac OS X or ``~/.bashrc`` on Ubuntu) and add a line similar to ``export _JAVA_OPTIONS="-Xms512m -Xmx4g"`` (or ``set _JAVA_OPTIONS=-Xms512m -Xmx4g`` on Windows)."

freeman-lab:
Thanks for the contribution and discussion @vjlbym and @GrantRVD ! I left a comment with a suggested change to one sentence that hopefully integrates the WIndows-specific instructions into this one PR, avoiding the need for another one. Let me know if it looks ok to you!

vjlbym:
Yeah, this looks OK to me. Thanks @freeman-lab and @GrantRVD

freeman-lab:
Ok great, thanks all.

PR #175, opened 2015-04-21
Select by index mask

jwittenbach:
Implements feature request #156.

Added an optional `returnMask` keyword argument to `Series.selectByIndex` which returns the mask used during the selection as an additional return value. Default value is `False`.


freeman-lab:
This LGTM, thanks @jwittenbach !

PR #176, opened 2015-04-22
Add Transpose to RowMatrix

rhofour:
This implements a (reasonably) efficient transpose for the RowMatrix class.


freeman-lab:
Thanks for the effort @rhofour , and sorry for the delay! Moving forward with the changes detailed in https://github.com/thunder-project/thunder/pull/243, thunder's data arrays when used with `spark` will be backed by `bolt` arrays, which includes a highly optimized block `transpose`. So I'll close this PR, though it's definitely a cool implementation, and could be a useful reference for other work!

PR #178, opened 2015-04-25
Add note to times to make its usage more clear.

rhofour:


freeman-lab:
Great clarification, thanks!

PR #181, opened 2015-05-03
Add RandomData example data generation.

rhofour:
This allows the generation of matrices where every element
comes from a standard normal distribution.

### DIFF
--- a/python/thunder/utils/datasets.py
+++ b/python/thunder/utils/datasets.py
@@ -97,6 +97,26 @@ class FactorAnalysisData(DataSets):
         else:
             return data
 
+class RandomData(DataSets):
+
+    def generate(self, nrows=50, ncols=50, npartitions=10, seed=None):
+        """
+        Generate a matrix where every element is i.i.d. and drawn from a
+        standard normal distribution
+
+        Parameters
+        ----------
+        nrows : int, optional, default = 50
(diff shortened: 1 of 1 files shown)


python/thunder/utils/datasets.py
```diff
@@ -97,6 +97,26 @@ def generate(self, q=1, p=3, nrows=50, npartitions=10, sigmas=None, seed=None):
...
+    def generate(self, nrows=50, ncols=50, npartitions=10, seed=None):
+        """
+        Generate a matrix where every element is i.i.d. and drawn from a
+        standard normal distribution
+
+        Parameters
+        ----------
+        nrows : int, optional, default = 50
+          Number of columns in the generated matrix
+        nrows : int, optional, default = 50
```
freeman-lab: Add line break here

freeman-lab:
Nice addition, thanks @rhofour ! I just had one formatting nit.

python/thunder/utils/datasets.py
freeman-lab: Like this:

```
nrows : int, optional, default = 50
    Number of columns in the generated matrix

nrows : int, optional, default = 50
    Number of rows in the generated matrix
```

freeman-lab:
LGTM, thanks!

PR #183, opened 2015-05-05
Made RandomData generate lazily.

okahn:


freeman-lab:
@okahn thanks for this, nice patch! LGTM

PR #195, opened 2015-06-12
Correlate bugfix

### LINKED ISSUE #194: Series.correlate() fails when index is multi-level
`Series.correlate(x)` checks that the length of `x` is the same as the length of the `Series.index`. However it does this through `numpy.size(self.index)`. However, in the case that the argument to `size` is more than one-dimensional, it returns the total number of elements in the array, not the length of the first dimension, as would be correct for a multi-index.

Simple solution: `size(self.index)` should be replaced with `self.index.shape[0]`.

jwittenbach:
Fixes #194


freeman-lab:
Something weird here, why is it pulling in a bunch of commits related to regression?

freeman-lab:
As for the fix, I'm not sure I'm crazy about it because the index itself might not be a numpy array (and thus won't have `shape` as a method). As examples, the index can sometimes be a single string (e.g. `mean`), sometimes a single number (e.g. 0), and sometimes a list (but not a numpy array). `size` works in all those cases, but I get that it fails for multi-index.

jwittenbach:
I'm not really sure what's wrong with the commit history. If you look at the diff, you can see that it hasn't actually changed anything other than the two lines in question...

I see what you mean about the fix being less than ideal. One alternative would be to just use the Python built-in `len` function. This would mean that a lone string would no longer work though, as `len("text")` will return `4`. We could require that such cases wrap the string in a list (e.g. `["text"]`). To me, this seems to be perhaps the least inelegant solution.

freeman-lab:
@jwittenbach closing this PR, as moving forward with 1.0.0 this will be better fixed by checking against the data's `shape` attribute, I just committed a change with a fix and added a unit test.

PR #198, opened 2015-06-24
Add in spark-ec2-git-repo and spark-ec2-git-branch default parameters

andrewgiessel:
New in 1.4.0, they changed the way they did AMI discovery.  See:  https://github.com/apache/spark/commit/b884daa58084d4f42e2318894067565b94e07f9d

Essentially, it relies on a default parameter that we didn’t have.  I followed the examples from the other parameters and added the default in line. 

I am spinning up clusters now, it got further than it did before, but they’re not up yet.  If they don’t spin up, I’ll post a comment.  @freeman-lab

### DIFF
--- a/python/thunder/utils/ec2.py
+++ b/python/thunder/utils/ec2.py
@@ -452,7 +452,11 @@ if __name__ == "__main__":
         parser.add_option("--placement-group", type="string", default=None,
                           help="Which placement group to try and launch instances into. Assumes placement "
                                "group is already created.")
-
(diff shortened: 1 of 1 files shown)


andrewgiessel:
After this second commit (adding an option for private-ips) this worked for me with `-s 1`.  Spinning up 8 now and we'll see.

freeman-lab:
This looks great, thanks for the fix @andrewgiessel !

PR #208, opened 2015-07-19
Replace scala dependency with binaryRecords

freeman-lab:
This a long overdue PR that eliminates all scala dependencies completely from Thunder. The only existing dependency was due to a binary loading scheme, but came with a lot of excess baggage: a non-standard folder hierarchy, the need to manage and bundle a jar file, issues with incompatible Hadoop versions, and jar argument handling during launch. Since the necessary loading scheme is now in Spark core as `binaryRecords`, we can use that and remove the Scala dependencies.

The key change is fairly minor but the reorganization is significant.


andrewgiessel:
bravo!

nehalecky:
Impressive work, congrats!

PR #210, opened 2015-07-19
remove dead packages in setup.py

andrewgiessel:
@freeman-lab `python setup.py install` fails on master due to these two lines which don’t exist in the source tree anymore.  Seems to work if I drop them- is there anything else that needs to change in this file?  I’ll tack it on if so.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,6 @@ setup(
               'thunder.extraction.block.methods',
               'thunder.extraction.feature.methods',
               'thunder.registration',
-              'thunder.registration.regmethods',
-              'thunder.lib',
               'thunder.rdds',
               'thunder.rdds.fileio',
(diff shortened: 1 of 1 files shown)


andrewgiessel:
yup we need to add registration.methods too at least.  I'll ping again when I think this is done

andrewgiessel:
Ok, I think it might be done.  @freeman-lab your recent refactoring out of scala means that I can use the spark installed by brew! :clap: :clap: :clap: :+1:

freeman-lab:
Good catch, thanks @andrewgiessel , I just botched the name during the refactor. Yes, I think this refactor was all around a big win, dealing with otherwise irrelevant Hadoop versioning issues purely for that one bit of functionality was really annoying.

Issue #209, opened 2015-07-19
Support for Python 3.4


andrewgiessel:
Thunder should have support for Python 3.4.

auvipy:
+python 3.5

freeman-lab:
Thanks @auvipy , this will definitely happening as part of the major refactoring currently underway, and almost finished!

https://github.com/thunder-project/thunder/pull/237

jwittenbach:
Added support for Python 3.4 in #247 and Python 3.5 is #270

Issue #141, opened 2015-03-24
Compatibility with Python 2.6


freeman-lab:
We should aim to ensure Thunder is compatible with Python 2.6 as much as possible without changing any functionality. This Travis report shows the current failures on a 2.6 build:

https://travis-ci.org/thunder-project/thunder/jobs/55596415

Most involve the use of `OrderedDicts`, and some unit test features (`skipIf` and `assert_is_none`). Both should be quite easy to fix either by changing our usage or suggesting workarounds (e.g., we could recommend users `pip install` a backport of `OrderedDict` as described [here](http://stackoverflow.com/questions/1617078/ordereddict-for-older-versions-of-python) if we detect they are running on 2.6, and include that install in the Travis setup).

If and when any more significant pieces of functionality require Python 2.7+, we can detect and skip those unit tests when running on a 2.6 environment.

PR #213, opened 2015-07-23
adding python versions to travis

andrewgiessel:
Initial work to explicitly test other Python versions (2.6, 3.3, and 3.4) via Travis.  Opening a PR to see if the build passes.

@freeman-lab I took out the if/else here that downloaded a different miniconda.  Is it needed?  For instance [here](https://github.com/python-visualization/folium/blob/master/.travis.yml), although the versions are passed through in a different way, they use the same miniconda script for all 3 python versions.

I am happy to change it back if it's installing something specific.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,18 @@
 language: python
 
 python:
+  - "2.6"
   - "2.7"
-
+  - "3.3"
+  - "3.4"
+  
 jdk:
   - openjdk7
 
 before_install:
   # install miniconda
   - sudo apt-get update
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      wget http://repo.continuum.io/miniconda/Miniconda-3.7.3-Linux-x86_64.sh -O miniconda.sh;
-    else
-      wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh;
-    fi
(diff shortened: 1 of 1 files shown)


andrewgiessel:
Ya, this kicks off explicit tests in all 4 versions.  3.3 and 3.4 failed pretty quickly.  2.7 passed.  2.6 is pending as I type this.  If there is no penalty to testing multiple travis builds, we can merge this, otherwise, it can be a guide post for us to think about compatibility changes (see #209 #141).

andrewgiessel:
2.6 also failed, but with much fewer problems than 3.x.  Many tests pass in fact and some of the issue are only in the tests themselves!

Here is a summary of the things to do:

1) install back port of `OrderedDict` with `pip install ordereddict` This is 2.4-2.7 compatible, and so should work as a drop-in replacement.  All `OrderedDict` imports must be from this module instead of `collections`. Probably one of the things that needs to be differentially imported in python3.

2)  `AttributeError: 'module' object has no attribute 'skipIf'`  this is a unittest decorator.
3) nose error:  `ImportError: cannot import name assert_is_none` this is a derived import in nose from unittest. 

I believe these can be fixed with `pip install unittest2`, which is a back port of 2.7+ specific `unittest` features.  Then, the missing features need to be imported from `unittest2` directly.  This is a nice clean import which is compatible up to python 3.4.

jwittenbach:
This is now implemented with PyTest. Closing.

Issue #193, opened 2015-06-11
add axis method to tile


freeman-lab:
It would be useful if `Colorize.tile` accepted an `axis` argument, to indicate which axis along which the individual images are indexed. For example, given an array `img` with shape `[10, 25, 5]`, calling

```
tile(img, axis=0)
```
would generate 10 different images each of shape `[25,5]`, whereas
```
tile(img, axis=2)
```
would generates 5 different images each of shape `[10,25]`

thanks for @tomsains for suggesting!

tomsains:
Hoping to take a stab at this. Im Just a little unsure about how 'ax' is defined in Colorize.image, should I use that same method?

jwittenbach:
@tomsains

###Colorize.image
In matplotlib, each 'figure' can have multiple 'axes' that you can draw on (think subplots). So in `Colorize.image`, that lets you pass in a matplotlib 'axis' object so that you can draw the image onto a subplot in a pre-existing figure if you want.

###Colorize.tile
However, for `Colorize.tile`, you almost always just want a new figure, so we don't need to worry about that here. For this function, a grid of images is passed in as a multi-dimensional array, and the `axis` argument should indicate which dimension in that array indexes the different images -- right now it assumes that it's the first.

freeman-lab:
Yup, all correct, and to avoid what would otherwise be pure chaos we use `ax` to refer to the matplotlib style axis in `tile`, and we can use `axis` in `image` to refer to the (array) axis. Trying to stick to matplotlib conventions, thought it's somewhat unavoidable confusion that the word axis refers to both things.

PR #214, opened 2015-07-24
add axis arg to Colorize.tile()

andrewgiessel:
Implements #193.  No tests for Colorize?


freeman-lab:
Nice fix, merging in! And yes, tests on Colorize defintiely needed but currently missing.

Issue #164, opened 2015-04-12
Documenting arguments for executable


logang:
The Thunder executable has useful additional arguments, like `-i` and `-n` for using IPython or the IPython notebook, but currently these are not listed when calling `thunder --help` because we only list the `pyspark` arguments. We should add these because they are useful and somewhat hidden.

PR #216, opened 2015-07-24
Enhance thunder exe script help string

andrewgiessel:
addresses #164


freeman-lab:
Thanks @andrewgiessel , merging this in so it's part of the legacy 0.6.0 branch release, but this will no longer be relevant moving forward with 1.0.0 as described herehttps://github.com/thunder-project/thunder/pull/243

Issue #162, opened 2015-04-10
Add explained variance output for PCA


freeman-lab:
PCA should return explained variance as one of its outputs

PR #217, opened 2015-07-24
add calculation of explained variance + ratio attributes to PCA

andrewgiessel:
Addresses #162.

@freeman-lab, can you please double check me here?  This needs a test.

### DIFF
--- a/thunder/factorization/pca.py
+++ b/thunder/factorization/pca.py
@@ -71,6 +71,11 @@ class PCA(object):
         self.latent = svd.s
         self.comps = svd.v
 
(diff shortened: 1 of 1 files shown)


andrewgiessel:
@j-friedrich thanks for the comment!  You're totally right.  The truncation of the components happens in `svd.py`, I guess I need to dig through there, too.

freeman-lab:
Closing this PR, as these factorization methods are moving to a separate repo, and this needs some further work anyway (it won't always be possible to compute explained variance ratio, if only a subset of components are estimated...).

PR #218, opened 2015-07-28
added offset kwarg for timeseries.py

d-v-b:
The `TimeSeries.normalize()` method currently uses the following expression for subtraction / division by baseline: 
`(y - b) / (b + 0.1)`

This PR exposes the value added in the denominator of the above expression as a kwarg, termed `offset`, to `TimeSeries.normalize()`. `offset` defaults to 0.1. 

normalization is now down thus:

`(y - b) / (b + offset)`

Example usage: `data.normalize(method='linear', offset=.5)`

### DIFF
--- a/thunder/rdds/timeseries.py
+++ b/thunder/rdds/timeseries.py
@@ -278,7 +278,7 @@ class TimeSeries(Series):
 
         return self.applyValues(func, keepIndex=True)
 
-    def normalize(self, baseline='percentile', window=None, perc=20):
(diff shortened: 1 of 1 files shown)


andrewgiessel:
Nice and easy and a good idea to expose it, thanks for the PR @d-v-b !

Should there be limits on the values that offset can take?  (i.e., non-negative?)

d-v-b:
I can't think of any obvious restrictions on the offset value -- someone might have negative-valued data, in which case they would need a negative offset to prevent dividing by 0...

Issue #152, opened 2015-04-04
Notebook pylab error on EC2


freeman-lab:
After upgrading the Python and IPython versions on our standard EC2 deployment, the following configuration we were using in the `setup-notebook-ec2` script is causing an error:

```
echo 'export IPYTHON_OPTS="notebook --pylab inline"' >> /root/.bash_profile
```

The error appears as

```
[E 07:55:53.493 NotebookApp] Support for specifying --pylab on the command line has been removed.
[E 07:55:53.493 NotebookApp] Please use `%pylab inline` or `%matplotlib inline` in the notebook itself.
The `pylab` should be removed.
```

A temporary workaround is to edit that line inside `~/.bash_profile` on the EC2 cluster after the notebook setup is complete, removing the `--pylab inline`.

freeman-lab:
Fixed by https://github.com/thunder-project/thunder/commit/039cd4e267d3cbc00b002e6944631ec6120131f1

PR #220, opened 2015-08-03
Fixed related to issue #152: "Notebook pylab error on ec2"

nerduno:
Issue #152 was fixed in setup-notebook-ec2, but not setup-notebok-ec2-sshtunnel.  This quick pull request fixes the same bug in setup-noteboke-ec2-sshtunnel.

-Aaron


freeman-lab:
Nice catch @nerduno ! Didn't realize this was still there, thanks. Merging in.

PR #221, opened 2015-08-04
fixed indexing issue in detrending

d-v-b:
`TimeSeries.detrend(data)` can return values outside of the range of the values in `data`. This is because the `TimeSeries.detrend()` method uses the wrong indices when calling `numpy.polyfit(vector, indices)`: specifically, the indices are shifted up by one. This PR fixes this bug by shifting the indices down by one.

### DIFF
--- a/test/test_timeseries.py
+++ b/test/test_timeseries.py
@@ -64,7 +64,7 @@ class TestTimeSeriesMethods(TimeSeriesTestCase):
         rdd = self.sc.parallelize([(0, array([1, 2, 3, 4, 5]))])
         data = TimeSeries(rdd).detrend('linear')
         # detrending linearly increasing data should yield all 0s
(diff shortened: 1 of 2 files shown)


d-v-b:
this fails here:
```
    def test_detrend(self):
        rdd = self.sc.parallelize([(0, array([1, 2, 3, 4, 5]))])
        data = TimeSeries(rdd).detrend('linear')
        # detrending linearly increasing data should yield all 0s
        assert(allclose(data.first()[1], array([0, 0, 0, 0, 0])))
```

but detrending `[1,2,3,4,5]` should return all 1s, not all 0s, so the test code also needs to be fixed

d-v-b:
Fixed the test code, now all tests pass

freeman-lab:
Sorry for the delay on this, merging in here, also added the same fix to the 1.0.0 branch.

PR #229, opened 2015-08-21
Update to create a rdd from a list of images

bald6354:
Saves the user from having to copy files to a single local directory if their data is spread across a file system and they already have a list

### DIFF
--- a/thunder/rdds/fileio/readers.py
+++ b/thunder/rdds/fileio/readers.py
@@ -186,9 +186,11 @@ class LocalFSParallelReader(object):
 
         Returns RDD of <integer file index, string buffer> k/v pairs.
         """
-        absPath = self.uriToPath(dataPath)
-        filePaths = self.listFiles(absPath, ext=ext, startIdx=startIdx, stopIdx=stopIdx, recursive=recursive)
-
+        if not hasattr(dataPath, '__iter__'):
+            absPath = self.uriToPath(dataPath)
+            filePaths = self.listFiles(absPath, ext=ext, startIdx=startIdx, stopIdx=stopIdx, recursive=recursive)
+        else:
+            filePaths = [filePath for filePath in dataPath]
         lfilepaths = len(filePaths)
         self.lastNRecs = lfilepaths
         npartitions = min(npartitions, lfilepaths) if npartitions else lfilepaths
@@ -626,8 +628,11 @@ SCHEMAS_TO_FILEREADERS = {
 def getByScheme(dataPath, lookup, default):
     """Helper function used by get*ForPath().
     """
-    parseresult = urlparse.urlparse(dataPath)
-    clazz = lookup.get(parseresult.scheme, default)
+    if hasattr(dataPath, '__iter__'):
+        clazz = LocalFSParallelReader
+    else:
+        parseresult = urlparse.urlparse(dataPath)
+        clazz = lookup.get(parseresult.scheme, default)
     if clazz is None:
         raise NotImplementedError("No implementation for scheme " + parseresult.scheme)
     return clazz


thunder/rdds/fileio/readers.py
```diff
@@ -626,8 +628,11 @@ def mode(self):
 def getByScheme(dataPath, lookup, default):
     """Helper function used by get*ForPath().
     """
-    parseresult = urlparse.urlparse(dataPath)
-    clazz = lookup.get(parseresult.scheme, default)
+    if hasattr(dataPath, '__iter__'):
```
freeman-lab: This implementation only works for a local file system, not S3, right? We can limit this PR to the local case, but would the S3 case be useful to have as well?

freeman-lab:
Thanks @bald6354 , this will be really nice to have! I'm almost done with a fairly major refactoring of Thunder's IO that will change how some of this works under the hood, but we can probably merge this in for now, and I'll make sure the new version supports the same functionality (I'll ping you on my PR).

For now, can you tell why the tests might be failing? Did you try running them locally?

bald6354:
I saw that it didn't pass. Idk why. We did not run them local. It shouldn't have impacted current functionality, but we overlooked something. I will look at it more next week. I think it would also be useful for s3. We have already started using this for local storage.
On Aug 21, 2015 7:35 PM, Jeremy Freeman <notifications@github.com> wrote:Thanks @bald6354 , this will be really nice to have! I'm almost done with a fairly major refactoring of Thunder's IO that will change how some of this works under the hood, but we can probably merge this in for now, and I'll make sure the new version supports the same functionality (I'll ping you on my PR).

For now, can you tell why the tests might be failing? Did you try running them locally?

bald6354:
The test also fail locally as well.  The error is from test_fromOCP in the test_imagesloader.  This should not occur based on the edits I submitted.  I pulled current project from git and got the same error in the test when I ran without the submitted changes.

freeman-lab:
@bald6354 awesome, thanks for tracking this down! This turns out to be a frustratingly flaky test that depends on a remote data source, I've disabled that test on master and everything is passing, so I'll go ahead and merge this, as that was the only test failure, and will look into a better solution in the meantime.

Thanks for the contribution! If you want to add the S3 version I'd hold out and wait for my newly refactored IO PR (coming soon...).

bald6354:
I am seeing a performance issue in constantly loading lots of TIFFs into spark. I am starting to look into storing the tiffs in a single file in HDFS and loading the single large file into thunder. Will the new IO address this issue? Let me know how to help. Thanks!

freeman-lab:
Can you say more what you mean by performance issue? What kind of environment is this, on EC2 loading from S3? Or some other filesystem?

Could be a number of things, one being the level of parallelism (which you can try changing with the `npartitions` argument). At the core, for the image loading, at least on S3, we're opening up one connection to S3 to per partition, and then loading one image in a partition at a time. This is unlikely to change with the new IO codebase, which is more about cleaning up the API and making it easier to add new formats.

Depending on the number and size of the images, some kind of batched IO could definitely be better (e.g. loading fewer files with several images in each, one per partition). But I'd be a little surprised if putting all images into a single file and loading that helped, because you lose any parallel IO.

Ultimately all of these questions are empirical, so any benchmarks would be incredibly welcome in trying to find better solutions! =) And if you describe your data a bit more (or provide a dimension-matched sample) we can try to do the same.

freeman-lab:
Ah, realizing from above your might be doing all the loading on a local file system, and then doing parallel computation from then onward?

bald6354:
Yes.  Loading the data is more than half of our processing time.  The files are ~1MB each (512x512).  We load hundreds of them for processing.  Our thought is to aggregate TIFFs for an area into larger files to take advantage of HDFS.

Issue #244, opened 2016-01-20
Inclusion/Exclusion on Series.between


jwittenbach:
When using `Series.between(start, stop)`, the behavior at the endpoints is currently dictated by a kwarg called `inclusive` that can be set to `True` to include the endpoints, or `False` to exclude them. Perhaps a better approach would be to always use the NumPy default of inclusion on the start index and exclusion on the stop index with no possibility of changing this. This one pattern can cover all cases and is what most people would probably expect anyway.

PR #245, opened 2016-01-20
update to Series.between inclusion behavior

jwittenbach:
Implements #244


freeman-lab:
thanks @jwittenbach ! assuming tests pass, this LGTM

PR #243, opened 2016-01-08
1.0.0

freeman-lab:
This PR tracks what will be a 1.0.0 release. It reflects a major refactoring, including many breaking API changes, thus the 1.0.0 label! Below are some of the primary changes (nearly all are complete):

- [x] support for both local analysis (w/ numpy) and distributed analysis (w/ Spark) via a common API
- [x] a design that can support a variety of computational backends
- [x] py.test for unit tests
- [x] use S3 for example data so it's not included in the repo
- [x] no more executables or standalone scripts to simplify deployment and installation
- [x] more consistent and python-friendly style conventions
- [x] updated documentation
- [x] python 3 compatibility

Another major change is breaking out functionality into a more modular collection of smaller libraries that can be independently versioned, updated, and tested. In this reorganization, `thunder` will be responsible only for the core operations of loading and manipulating image and time series, but can be composed and used alongside with several additional libraries, which themselves can be used for either local or distributed purposes.

Analytics
- [x] [`thunder-extraction`](https://github.com/thunder-project/thunder-extraction) - feature extraction from image sequence data
- [x] [`thunder-registration`](https://github.com/thunder-project/thunder-registration) - registration for image sequence data
- [x] [`thunder-factorization`](https://github.com/thunder-project/thunder-factorization) - matrix factorization methods
- [x] [`thunder-regression`](https://github.com/thunder-project/thunder-regression) - mass-univariate regression modeling

Utilities
- [x] [`bolt`](https://github.com/bolt-project/bolt) - ndarray-like interface for spark
- [x] [`station`](https://github.com/freeman-lab/station) - context manager for computational engines
- [x] [`checkist`](https://github.com/freeman-lab/checkist) - argument checking
- [x] [`showit`](https://github.com/freeman-lab/showit) - simple display of images and tiled images

PR #247, opened 2016-01-22
Python 2 and 3 compatibility

jwittenbach:
This PR address one of the goals of #243 -- making v1.0.0 compatible both Python 2 and 3

- [x] get unit tests to run
   - [x] base
   - [x] blocks
   - [x] images
   - [x] images_io
   - [x] readers
   - [x] series
   - [x] series_io
   - [x] timerseries
- [x] update Travis to test both Python 2 and 3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,7 @@ numpy
 scipy
 scikit-image
 boto
+pillow
+six
 bolt-python >= 0.3.0
 checkist >= 1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,5 @@ numpy
 scipy
 scikit-image
 boto
-pillow
-six
 bolt-python >= 0.3.0
 checkist >= 1.0.0
--- a/test/test_series_io.py
+++ b/test/test_series_io.py
@@ -81,8 +81,7 @@ def test_from_text_skip(tmpdir):
 def test_from_binary(tmpdir, eng):
     a = arange(8, dtype='int16').reshape((4, 2))
     p = os.path.join(str(tmpdir), 'data.bin')
-    with open(p, 'w') as f:
-        f.write(a)
+    a.tofile(p)
     data = frombinary(p, shape=[4, 2], dtype='int16', engine=eng)
     assert allclose(data.shape, (4, 2))
     assert allclose(data.index, [0, 1])
@@ -94,8 +93,7 @@ def test_from_binary_skip(tmpdir, eng):
     v = [[0, i] for i in range(10)]
     a = array([kv[0] + kv[1] for kv in zip(k, v)], dtype='int16')
     p = os.path.join(str(tmpdir), 'data.bin')
-    with open(p, 'w') as f:
-        f.write(a)
+    a.tofile(p)
     data = frombinary(p, shape=[10, 2], dtype='int16', skip=1, engine=eng)
     assert allclose(data.shape, (10, 2))
     assert allclose(data.index, [0, 1])
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -1,5 +1,7 @@
 import logging
 from numpy import ndarray, arange, amax, amin, size, asarray, random, prod
+# Name changes between Python 2 and 3
+from six.moves import xrange
 
 from ..base import Data
 
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -165,7 +165,7 @@ class Images(Data):
         Execute a function on each image
         """
         if self.mode == 'spark':
-            self.values.tordd().map(lambda (k, v): (k[0], v)).foreach(func)
+            self.values.tordd().map(lambda kv: (kv[0][0], kv[1])).foreach(func)
         else:
             [func(kv) for kv in enumerate(self.values)]
 
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -1,7 +1,5 @@
 import logging
 from numpy import ndarray, arange, amax, amin, size, asarray, random, prod
-# Name changes between Python 2 and 3
-from six.moves import xrange
 
 from ..base import Data
 
@@ -320,8 +318,8 @@ class Images(Data):
         def roundup(a, b):
             return (a + b - 1) // b
 
-        slices = [slice(0, dims[i], factor[i]) for i in xrange(ndims)]
-        newdims = tuple([roundup(dims[i], factor[i]) for i in xrange(ndims)])
+        slices = [slice(0, dims[i], factor[i]) for i in range(ndims)]
+        newdims = tuple([roundup(dims[i], factor[i]) for i in range(ndims)])
 
         return self.map(lambda v: v[slices], dims=newdims)
 
--- a/thunder/images/readers.py
+++ b/thunder/images/readers.py
@@ -2,8 +2,6 @@ import itertools
 import checkist
 from io import BytesIO
 from numpy import frombuffer, prod, random, asarray, expand_dims
-# library reorganization between Python 2 and 3
-from six.moves import xrange
 
 from ..utils import check_spark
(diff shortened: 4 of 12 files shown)


jwittenbach:
@freeman-lab I think this is ready to go. A couple of change highlights:

1. Added `six` to the requirements. This package provides a level of abstraction for for accessing standard library functions/modules that changes names/locations between Python 2 and 3. Without this package, the code needed to make things compatible with Python 2 and 3 simultaneously would have been much much uglier.

2. There's a big difference between the way that Python 2 handles binary data (encoded as strings) and how Python 3 does it (a dedicated `bytes` object). When reading/writing text data to a file in binary mode (something that all of Thunder's readers/writers do), this means that, in Python 3, you now need to explicitly specify an encoding/decoding scheme. I have use `utf-8` everywhere.

3. While not necessarily a Python 2 vs 3 problem, I added `pillow` to the requirements as well. When testing my Python 3 changes, I made a minimal Conda environment, and the base version of SciPy does not pull Pillow along with it, but it is needed to use `scipy.mis.imread`. Thus, it seems best to explicitly put this in the `requirements.txt` file so anyone pip-installing Thunder doesn't end up without it.

test/test_series_io.py
```diff
@@ -81,8 +81,7 @@ def test_from_text_skip(tmpdir):
 def test_from_binary(tmpdir, eng):
     a = arange(8, dtype='int16').reshape((4, 2))
     p = os.path.join(str(tmpdir), 'data.bin')
-    with open(p, 'w') as f:
-        f.write(a)
+    a.tofile(p)
```
freeman-lab: was this just for shortening?

thunder/images/images.py
```diff
@@ -1,5 +1,7 @@
 import logging
 from numpy import ndarray, arange, amax, amin, size, asarray, random, prod
+# Name changes between Python 2 and 3
+from six.moves import xrange
```
freeman-lab: It seems more "forward looking" to me to just replace all the `xrange`s with `range` and then fix any residual uses of `range` for py2 where necessary (e.g. by wrapping in a list or tuple). Put another way, although `six` makes it easy, it seems backwards to essentially keep using a deprecated API.

thunder/images/readers.py
```diff
@@ -2,6 +2,8 @@
 import checkist
 from io import BytesIO
 from numpy import frombuffer, prod, random, asarray, expand_dims
+# library reorganization between Python 2 and 3
```
freeman-lab: same comment as above about `xrange`->`range`

thunder/series/readers.py
```diff
@@ -309,16 +320,22 @@ def frombinary(path, ext='bin', conf='conf.json', dtype=None, shape=None, skip=0
...
+        def switch(record):
+            ary, idx = record
+            return (idx,), ary
+
+        rdd = raw.zipWithIndex().map(switch)
 
         if shape and len(shape) > 2:
             expand = lambda k: unravel_index(k[0], shape[0:-1])
-            rdd = rdd.map(lambda (k, v): (expand(k), v))
+            rdd = rdd.map(lambda kv: (expand(kv[0]), kv[1]))
```
freeman-lab: sometimes you fixed this by writing a new function, other times by expanding inside the lambda, was there a reason? sort of prefer the latter, and at least prefer consistency

thunder/series/series.py
```diff
@@ -5,6 +5,9 @@
     ravel, logical_not, max, min, unravel_index, prod, random, shape, \
     dot, outer, expand_dims, ScalarType, ndarray
 from bolt.utils import tupleize
+# naming changes between Python 2 and 3
+from six import string_types
+from six.moves import xrange
```
freeman-lab: see comment about `xrange`

requirements.txt
```diff
@@ -2,5 +2,7 @@ numpy
 scipy
 scikit-image
 boto
+pillow
```
freeman-lab: i'd rather not add this, and essentially make it optional by doing a check inside `frompng` and `topng`

freeman-lab:
Amazing work on this @jwittenbach ! And doesn't seem like it was as bad as I feared. I left a handful of suggestions and comments, all fairly minor, the main one involving `xrange`. My thinking was we want to be as forward-looking as possible with these changes, even if it's a bit more work.

The other major comment is re: `pillow`, we do depend on `scikit-image`, and that depends on `pillow`, so I'm confused why you found it necessary to specify? I've always found `pillow` to be a minefield as a dependency, so while I'm not surprised something is funny, I'd also love to not have to deal with it ourselves, and if that means making a couple things optional, that's fine.

requirements.txt
freeman-lab: On second thought, it should be required because of `scikit-image`!

test/test_series_io.py
jwittenbach: Nope, that doesn't work in Python 3. It comes to this issues of `bytes` and `string` objects being different in Python 3. In Python 3, opening a file in `w` mode means that you have to pass it a string. When you pass it an `ndarray` instead it calls `ndarray.tostring()` which, in Python 2 returns a string (that encodes the bytes). But in Python 3 it returns a `bytes` object instead, which the file writer does not accept.

In hindsight, opening the file in `wb` mode would make the old code work in both Python 2 and 3. But since the `ndarray.tofile()` method does the right thing on both versions and is more concise, I vote we just leave it.

thunder/images/images.py
jwittenbach: I can definitely see the logic in that. In choosing to implement it this way, my concern was with performance: if you use `range` everywhere, then people using Python 3 will reap the benefits of `range` returning a more performant generator object, while those in Python 2 won't get this boost. However, we're probably not actually using `range` anywhere that the object being generated is actually big enough that this is an issue, so I think see your concern for being more forward-looking as winning out. I'll switch to that solution.

thunder/series/readers.py
jwittenbach: My method here was the following -- I considered two cases:

1. When the key-value pair represented the standard Thunder record of `(index, ndarray)` I used the shorter but less interpretable `kv` inside a lambda.

2. When the key-value pair was not the standard Thunder record, I explicitly defined a new function so that the more explicitly argument names would help someone reading the code to understand what the objects being manipulated are.

To me, this makes sense, but I'll switch it if you're not swayed by my argument.

requirements.txt
jwittenbach: So we both agree now that it should be included?

thunder/images/images.py
freeman-lab: cool, great, totally see that reasoning, but also totally agree there are few places where that performance difference will matter.

test/test_series_io.py
freeman-lab: great, sounds good!

requirements.txt
freeman-lab: hm, no, i guess i meant, if it's included in `scikit-image`, we shouldn't need the explicit dependency right?

freeman-lab: a little confused by your failed install in that case...

jwittenbach: OK, my environment must have been borked. Just went back and tried it with a fresh environment and you're right -- `pillow` is included by `scikit-image` so that's all good. Incidentally, I also discovered that `six` also gets pulled in by `scikit-image`, so I'll remove that from the dependency list as well.

jwittenbach:
@freeman-lab Implemented all of the changes that we agreed on. The requirements list is beautifully short now :smile:

thunder/series/readers.py
freeman-lab: ok cool, that makes sense to me!

freeman-lab:
Awesome @jwittenbach  this looks good to me, merging in. Terrific work on this!

PR #249, opened 2016-02-04
keep singleton dimensions on reduce

freeman-lab:
Simple fix to make sure reduce doesn't eliminate singleton dimensions.


freeman-lab:
LGTM!

PR #250, opened 2016-02-15
1.0.0 local corr

sofroniewn:
Fix the implementation of local corr so it is now compatible with Thunder 1.0.0 and works in both spark and local mode. Still need to reenable the unit tests


freeman-lab:
nice, thanks @sofroniewn ! merging in, can add tests next.

freeman-lab:
fyi i think there might be a mismatch between your main github handle and whatever name/email was used on these commits (notice your icon isn't showing by the commits...)

sofroniewn:
ah yes - it warned me about that, will fix next time

PR #252, opened 2016-03-02
Blocks Update

jwittenbach:
This PR adds two features to the `Blocks` object:

1. Adds a `Blocks.map` operation for applying array-to-array operations on a  block by block basis.
2. Adds simple `local` mode support for blocks by considering the entire array to be a single block.


jwittenbach:
Ah shoot, opened this PR against the wrong branch. Closing...

PR #260, opened 2016-03-10
1.0.0 clean up code and docs for filtering on images

sofroniewn:
Previous filter operations on images expanded a scalar input parameter to two dimensions instead of three dimensions thereby treating 3D volumes as a collection of 2D images. This behaviour has been changed in 1.0.0, but the docs had not yet been updated to reflect it.

This PR updates the docs to indicate there is no more special handling of scalar input parameters, and removes some unnecessary code in the gaussian_filter method

### DIFF
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -438,7 +438,7 @@ class Images(Data):
         if not isinstance(neighborhood, int):
             raise ValueError("The neighborhood must be specified as an integer.")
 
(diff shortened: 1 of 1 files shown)


sofroniewn:
This PR now includes a fix of the localCorr so that it works in spark mode

jwittenbach:
LGTM :+1:

PR #262, opened 2016-03-22
commiting change

jwittenbach:
Small fix to how `Base._map` functions in `local` mode. This allows the mapped function to return a singleton value (that is not wrapped in an `ndarray`) and still function. The resulting `Base.values` will still be an `ndarray` with the correct number of dimensions.


jwittenbach:
Pretty minor change -- tests pass so I'm going to merge. @freeman-lab let me know if you have any issues with this.

PR #278, opened 2016-04-12
Remove reference to neighborhood in localcorr

sofroniewn:


sofroniewn:
Bug fix in localcorr - reference to undefined variable neighborhood now removed.
We need to add the localcorr unit tests back in as they would not have passed with this bug

freeman-lab:
@sofroniewn great catch, a separate PR restoring the unit tests would be terrific.

PR #279, opened 2016-04-13
allow tiffs with npages not dividing pageCount to be loaded

sofroniewn:
Currently if npages did not divide pageCount a value error was raised.

This PR changes the behaviour of fromtiff such that now all the data apart from the last few remaining pages will be loaded. These remaining pages are ignored.

### DIFF
--- a/thunder/images/readers.py
+++ b/thunder/images/readers.py
@@ -332,6 +332,7 @@ def fromtif(path, ext='tif', start=None, stop=None, recursive=False, nplanes=Non
 
     nplanes : positive integer, optional, default = None
         If passed, will cause single files to be subdivided into nplanes separate images.
+        If nplanes does not evenly divide the page count, the remaining pages will be discarded.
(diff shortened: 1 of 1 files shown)


boazmohar:
But will it raise a warning? 
We had this happen to us, and we used it as an indication that something went wrong with the acquisition.

sofroniewn:
Currently it does not raise a warning - I can amend the PR such that it does.

Alternatively we could keep the default behaviour the same and add some form of keyword argument that tells you to discard the remaining frames if you want.

Thoughts? Curious what @freeman-lab thinks is best too

boazmohar:
Thinking about at, I am still not sure what would happen if there is a missing volume in the middle of an acquisition, would the keys skip that time point? 
I think it would be helpful to have a keyword argument to discard the frames, but in a way that we would know what happened.
Maybe for each dropped volume a warning would display (In file xxx.tif dropped y frames) or the key would be skipped.

PR #285, opened 2016-04-20
Added dtype to images.map

### LINKED ISSUE #284: dtype propogation in images.map() operation
I am trying to do a map opperation that changes my value size and I know the resulting size.
Following the code I see that in thunder images ```map()``` exposes ```dims``` but not ```dtype```: 

``` def map(self, func, dims=None, with_keys=False): ```

It callse _map from the base class:

```return self._map(func, axis=0, value_shape=dims, with_keys=with_keys)```

without passing ```dtype```. 

```_map()``` does accept a ```dtype``` that defaults to ```None```:

```_map(self, func, axis=(0,), value_shape=None, dtype=None, with_keys=False):```

calling the bolt map operation always with ```None```:

``` mapped = self.values.map(expand, axis, value_shape, dtype, with_keys)```

In Bolt, if either ```dims``` which is now ```value_shape```  or ```dtype``` are ```None``` it will test the operation with a random array:

```python
if value_shape is None or dtype is None:
            # try to compute the size of each mapped element by applying func to a random array
            try:
                mapped = test_func(random.randn(*swapped.values.shape).astype(self.dtype))
            except Exception:
                first = swapped._rdd.first()
                if first:
                    # eval func on the first element
                    mapped = test_func(first[1])
            if value_shape is None:
                value_shape = mapped.shape
            if dtype is None:
                dtype = mapped.dtype
```
I will be happy to do a PR if anyone is willing to show me how (it will be my first)

Boaz

boazmohar:
fixes #284 
Only adds the functionality for images object not for series.


boazmohar:
Added dtype to series as well.
Something to consider in series.map()
There is currently no way to pass a shape of the resulting index without passing an index. It is an edge case, and I don't know if you want this to be implimented.
@freeman-lab What do you think?

jwittenbach:
@boazmohar this is looking great. Any chance you could add the same functionality to `Blocks.map` + a corresponding test? If we can get that in, then I'd be happy to merge this in.

boazmohar:
Sure I will do it later today.


Thanks,
Boaz On Mon, May 09, 2016 at 1:09 PM, Jason Wittenbach < notifications@github.com [notifications@github.com] > wrote:
@boazmohar [https://github.com/boazmohar] this is looking great. Any chance you could add the same functionality to Blocks.map + a corresponding test? If we can get that in, then I'd be happy to merge this in.

boazmohar:
@jwittenbach looking over the `blocks.py` code, there is a dtype passed there:
[blocks.py@map()](https://github.com/thunder-project/thunder/blob/master/thunder/blocks/blocks.py#L48)

I haven't seen a test for it, do you want me to add one?

jwittenbach:
@boazmohar ah, I should have looked more closely. Yeah, if you want to add a test, that would be perfect!

boazmohar:
@jwittenbach Done, could you please have a look, I want Aaron to use this fix as well.
Thanks!

jwittenbach:
@boazmohar awesome. Looks like there are some merge conflicts. This is probably because some other changes have been made in the meantime. If you do a `git pull upstream` from this branch on your local machine, then you it will try to grab those changes. Any files where those changes clash with yours will show up when you do `git status`. Inside those files it will show you the conflict, and you just need to manually resolve it. Then do another comit and push.

boazmohar:
@jwittenbach sorry about the mess, should work now.

jwittenbach:
Looks good to me. Thanks @boazmohar!

PR #290, opened 2016-04-26
Extending fromarray methods

### LINKED ISSUE #289: inconsistency in fromarray
When `Series.fromarray` is passed a `BoltArraySpark` object with no explicit `engine`, the resulting `Series` object has the same `mode` ("spark" vs "local") as the argument.

However, when the same function is passed a `Series` object, the resulting `Series` always defaults to have the `mode` set to "local".

jwittenbach:
closes #289 by extending Images and Series fromarray methods to handle Thunder objects


jwittenbach:
After chatting with @freeman-lab, decided to leave the functionality of `fromarray` as is. Closing.

PR #295, opened 2016-04-29
add decimation to Images

j-friedrich:
Decimates images to reduce data size (and imaging rate).
Sequentially averages N frames together, instead of merely taking every N-th frame, thus preserving better SNR. Corresponds to running mean filtering with window length N followed by subsampling by N.

### DIFF
--- a/test/test_images.py
+++ b/test/test_images.py
@@ -96,6 +96,13 @@ def test_subsample(eng):
     assert allclose(vals, truth)
 
 
+def test_decimate(eng):
+    data = fromlist(list(arange(24).reshape((4, 3, 2))), engine=eng)
+    vals = data.decimate(2).toarray()
+    truth = [arange(3, 9).reshape((3, 2)), arange(15, 21).reshape((3, 2))]
+    assert allclose(vals, truth)
+
+
 def test_median_filter_2d(eng):
     data = fromlist([arange(24).reshape((4, 6))], engine=eng)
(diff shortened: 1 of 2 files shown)


jwittenbach:
Another way to implement this would be to use `Images.map_as_series`. It might look something like this (haven't actually tested this, but the general idea should be right):

```python
def decimate(self, n):

   def decimate_block(block):
      return r_[ [mean(block[i:i+n].mean(axis=0)[np.newaxis] for i in arange(0, block.shape[0], n)]]

   new_length = int(np.ceil(self.shape[0]/n))
   return self.map_as_series(decimate_block, value_shape=new_length, dtype=np.float64)
```

I would be interested in know if there's a performance difference between these. I know that we've have troubles with `reduce` operations being slow in the past without lots of optimization.

j-friedrich:
Images.map_as_series transforms to blocks, which is slow. The main idea behind decimation is to quickly reduce the size of the data first and then transform to blocks to do source extraction for which temporally decimated data is the sweet spot between merely one summary image and the whole data.   

I expect a clear performance difference between these implementations, most dramatically if image 1 to n is on the first node, n+1 to 2n on the second, … thus requiring no shuffling between nodes at all. Of course it would be terrible if image i is on node i mod n. How get the images distributed in the first place?

freeman-lab:
@j-friedrich @jwittenbach it'd be really awesome to see performance numbers on this for a large test dataset, I'd say we just measure it and go with whichever implementation is faster.

PR #296, opened 2016-05-03
GDAL, Nan Calculations, and HDFS for loading images

stevevarner:
Adding changes to use GDAL for loading TIF images.

Adding changes to allow nan calculations (i.e. nanmean, nanmin, nanmax)

Adding changes to allow reading from HDFS when loading images.

### DIFF
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -1,11 +1,57 @@
 import logging
 from numpy import ndarray, arange, amax, amin, size, asarray, random, prod, \
-    apply_along_axis
+    apply_along_axis, nanmean, nanstd, nanmin, nanmax, nansum, nanmedian, inf, subtract
+	
 from itertools import product
 
 from ..base import Data
 from ..blocks.local import LocalBlocks
 
+class Dimensions(object):
+    """ Class for estimating and storing dimensions of data based on the keys """
+
+    def __init__(self, values=[], n=3):
+        self.min = tuple(map(lambda i: inf, range(0, n)))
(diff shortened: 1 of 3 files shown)


freeman-lab:
@stevevarner thanks for putting this together! A few top-level comments that will hopefully make it easier to incorporate these changes

- it'd be ideal if you could split this into three separate PRs for the three separate changes (GDAL, HDFS, nan)
- it'd be great to write new tests for each new piece of functionality, you can look at the existing tests as a guide 
- there's code in the commit that adds the `crop` method back in, but it's not mentioned in the PR or used in any of the other changes, maybe this was included by accident?
- the `gdal` related code seem to have some hardcoded paths, maybe that was included by accident too?

thanks!

techchrj:
@freeman-lab 

In reference to gdal:

you use FileFromMemBuffer() to create a virtual file and open it with the buffer

Here's the C API for it : 
http://gdal.org/cpl__vsi_8h.html#1ecb3a78729ecea4d2ce22065a605244

And here's an extract of a test in the Python autotest suite. 

    content = open('data/byte.tif', mode='rb').read()

    # Create in-memory file and initialize it with the content 
    gdal.FileFromMemBuffer('/vsimem/tiffinmem', content)

    # Open the in-memory file
    ds = gdal.Open('/vsimem/tiffinmem')

    # do something with the dataset

    # Close the dataset
    ds = None

    # Release memory associated to the in-memory file
    gdal.Unlink('/vsimem/tiffinmem')

PR #299, opened 2016-05-04
GDAL library for reading TIF files

stevevarner:
Adding changes to use GDAL for loading TIF images.


d-v-b:
@stevevarner what's the motivation for using GDAL over `tifffile.py`?

stevevarner:
1.  GDAL handles 32bit floating point TIF files properly.
2.  GDAL is widely used across the community 
3.  GDAL is regularly maintained and updated.

PR #298, opened 2016-05-04
NaN values in images

stevevarner:
Adding changes to allow nan calculations (i.e. nanmean, nanmin, nanmax)

PR #305, opened 2016-05-18
NaN handeling

boazmohar:
Continued work from #298
Removed the Dimensions class.
Added the right wrapping for local mode.
Added tests.
Only in Images so far.


boazmohar:
@freeman-lab I think this is it for NaNs, could you go over the code and tell me what you think?
Thanks!

PR #310, opened 2016-05-19
dtype not always of class numpy.dtype

boazmohar:
forced dtype property to be of class np.dtype so str(dtype) would give correct output


boazmohar:
Addresses #310

freeman-lab:
LGTM! Thanks @boazmohar

PR #311, opened 2016-05-20
Blocking scheme update

### LINKED ISSUE #307: block sizing
After extensive benchmarking with @jwittenbach , we realized a few issues with how block/chunk sizes are currently computed, and in particular found that the default value of `150` is too large. We came up with a plan that should significantly improve performance.

For the following functions:
- `images.toblocks()`
- `images.toseries()`
- `images.map_as_series()`

The argument currently known as `block_size` or `size` should be consistently named `chunk_size`. If given a string, it will be treated as `kb` **not** `mb`, so should be documented as such. The default values should be `auto` where `auto` tries to compute a `chunk_size` that will result in `~100mb` blocks, where a block includes all dimensions, whereas a chunk is defined only in image space.

For example, for `(9000,512,512)` data, the empirically optimal `chunk_size` is around `10kb`, which results in blocks of size `(9000,9,512)` which is around `100mb`.

For the time being, manually specifying sizes much lower than the current default could significantly improve performance for these operations.

cc @d-v-b

jwittenbach:
Closes #307

### DIFF
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -61,16 +61,17 @@ class Images(Data):
         if self.mode == 'spark':
             return self.values.tordd().values().first()
 
-    def toblocks(self, size='150', padding=None):
+    def toblocks(self, chunk_size='auto', padding=None):
         """
         Convert to blocks which represent subdivisions of the images data.
 
         Parameters
         ----------
-        size : str, or tuple of block size per dimension,
-            String interpreted as memory size (in megabytes, e.g. '64').
+        chunk_size : str or tuple, size of image chunk used during conversion, default = 'auto'
+            String interpreted as memory size (in kilobytes, e.g. '64').
+            The exception is the string 'auto'. In spark mode, 'auto' will choose a chunk size to make the
+            resulting blocks ~100 MB in size. In local mode, 'auto' will create a single block.
             Tuple of ints interpreted as 'pixels per dimension'.
-            Only valid in spark mode.
 
         padding : tuple or int
             Amount of padding along each dimensions for blocks. If an int, then
@@ -80,17 +81,18 @@ class Images(Data):
         from thunder.blocks.local import LocalChunks
 
         if self.mode == 'spark':
-            chunks = self.values.chunk(size, padding=padding).keys_to_values((0,))
+            if chunk_size is 'auto':
+                chunk_size = str(int(100000.0/self.shape[0]))
+            chunks = self.values.chunk(chunk_size, padding=padding).keys_to_values((0,))
 
         if self.mode == 'local':
-            if isinstance(size, str):
-                raise ValueError("block size must be a tuple for local mode")
-            plan = (self.shape[0],) + tuple(size)
-            chunks = LocalChunks.chunk(self.values, plan, padding=padding)
+            if chunk_size is 'auto':
+                chunk_size = self.shape[1:]
+            chunks = LocalChunks.chunk(self.values, chunk_size, padding=padding)
 
         return Blocks(chunks)
 
-    def toseries(self, size='150'):
+    def toseries(self, chunk_size='auto'):
         """
         Converts to series data.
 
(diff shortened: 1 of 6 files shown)


thunder/images/images.py
```diff
@@ -79,35 +80,44 @@ def toblocks(self, size='150', padding=None):
...
         This method is equivalent to images.toblocks(size).toSeries().
 
         Parameters
         ----------
-        size : string memory size, optional, default = '150M'
-            String interpreted as memory size (e.g. '64M').
+        chunk_size : str or tuple, size of image chunk used during conversion, default = 'auto'
+            String interpreted as memory size (in kilobytes, e.g. '64').
+            The exception is the string 'auto', which will choose a chunk size to make the
+            resulting blocks ~100 MB in size. Tuple of ints interpreted as 'pixels per dimension'.
```
freeman-lab: Not obvious what a `block` is, maybe say `to make the resulting intermediate objects ~100MB in size`. Will still be confusing that a specification of `1000` is very different than what you get with `auto`, not sure what to do about that...

thunder/images/images.py
```diff
@@ -553,11 +563,13 @@ def map_as_series(self, func, value_size=None, dtype=None, block_size='150'):
             dtype of one-dimensional ndarray resulting from application of func.
             If not supplied it will be automatically inferred for an extra computational cost.
 
-        block_size : str, or tuple of block size per dimension, optional, default = '150'
-            String interpreted as memory size (in megabytes e.g. '64'). Tuple of
-            ints interpreted as 'pixels per dimension'.
+        chunk_size : str or tuple, size of image chunk used during conversion, default = 'auto'
+            String interpreted as memory size (in kilobytes, e.g. '64').
+            The exception is the string 'auto', which will choose a chunk size to make the
+            resulting blocks ~100 MB in size. Tuple of ints interpreted as 'pixels per dimension'.
```
freeman-lab: same comment as above

thunder/images/images.py
```diff
@@ -79,35 +80,44 @@ def toblocks(self, size='150', padding=None):
         from thunder.blocks.blocks import Blocks
         from thunder.blocks.local import LocalChunks
 
+        if chunk_size is 'auto':
+            chunk_size = str(int(1000000.0/self.shape[0]))
```
freeman-lab: what's the logic here? you're only using the first dimension why?

freeman-lab:
@jwittenbach looking good, left a couple comments. Didn't totally follow the automatic size calculation.  And it's slightly confusing for us to say that a size of of `auto` means we'll pick a `chunk_size` to make the intermediate objects `1000` kb, but that's then very different than what you'd get if you specify a `chunk_size` of `1000` kb. Not sure there's a better way though.

thunder/images/images.py
jwittenbach: Yeah, let's make sure we get this right :)

I'm using the following terminology:
a "chunk" is a subset of (x, y, z)
a "block" is (t, x, y, z), where you have an (x, y, z) chunk and all time points

The block is the intermediate object in `Images.toseries` and `Images.map_as_series`. Bolt works with chunks, so if we want the blocks to be of size S=100MB when you choose `auto`, then the chunk that we should use is 100MB/len(t). That's the computation that's being done in line `84`.

Does that make sense?

freeman-lab: Ah ah, ok great, just had to think through it, yeah that totally makes sense. Shouldn't the equation take `dtype` into account though?

And before merging, let's do some sanity sets on real, large datasets to make sure the numbers come out as expected. Maybe you've already done that.

freeman-lab: Nevermind about `dtype`, forgot that was handled inside `bolt`. Would still be good to see the real use case confirmation.

freeman-lab: And I think you have an extra `0` should be `100000`

jwittenbach: Yep, removed the extra '0'

jwittenbach:
@freeman-lab I think this is ready to go now, though it might be good if you took one last look at it :)

freeman-lab:
looks good to me!

PR #313, opened 2016-05-22
tiffs with npages not dividing by pageCount will give a warning instead an error

boazmohar:
Alternative implementation to #279 
Thinking of adding a flag to `fromtif()` that would default to the current behavior of rising an error.
Also I am sending the warning using the `warning` module, I saw another way using the spark logger:
```python 
log4jLogger = sc._jvm.org.apache.log4j
LOGGER = log4jLogger.LogManager.getLogger(__name__)
LOGGER.warning("number of planes...")
```
@sofroniewn @freeman-lab What do you guys think?

### DIFF
--- a/test/test_images_io.py
+++ b/test/test_images_io.py
@@ -114,6 +114,16 @@ def test_from_tif_multi_planes(eng):
     assert [x.sum() for x in data.toarray()] == [1140006, 1119161, 1098917]
 
 
+def test_from_tif_multi_planes_discard_extra(eng):
+    path = os.path.join(resources, 'multilayer_tif', 'dotdotdot_lzw.tif')
+    data = fromtif(path, nplanes=2, engine=eng, discard_extra=True)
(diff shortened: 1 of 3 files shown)


freeman-lab:
Why don't we forget the warning stuff, because doing warnings on the workers well is hard, and instead add a keyword argument like `discard_extra` that will discard extra planes instead of giving an error (as we do now). Ok if you want at the same time to use our own warnings to log the file name.

freeman-lab:
Warnings should be done like this: 
https://github.com/thunder-project/thunder/blob/0f2e29fbe106a7b50dffeb6f056aba4c107b82a7/thunder/utils.py#L4

boazmohar:
@freeman-lab Thanks for the comments!
I think it is ready to merge.

freeman-lab:
Excellent, looks good to me @boazmohar , merging in!

PR #314, opened 2016-05-22
Checking py4j

### LINKED ISSUE #306: spark tests on travis
It appears that the travis CI tests may not in fact be running against Spark via the `--engine` flag, even though the same definitely works locally. We need to investigate this!

I made a branch that deliberately fails a test and includes a print statement showing the current "engine". The same thing run locally shows a `SparkContext` as the engine.

https://travis-ci.org/thunder-project/thunder/jobs/131200190

boazmohar:
might fix #306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ install:
   - source activate test
   - pip install .
   - pip install -r requirements-dev.txt
+  - pip install py4j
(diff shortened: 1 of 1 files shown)


boazmohar:
@freeman-lab I think it worked

PR #315, opened 2016-05-22
fix spark test on travis due to lack of py4j

### LINKED ISSUE #306: spark tests on travis

boazmohar:
fixes #306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ install:
   - conda create --yes -q -n test python=$TRAVIS_PYTHON_VERSION pip numpy scipy scikit-learn scikit-image matplotlib pytest boto
   - source activate test
   - pip install .
+  - pip install py4j
   - pip install -r requirements-dev.txt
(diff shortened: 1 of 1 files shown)


freeman-lab:
Again, awesome for figuring this out! Do you mind trying one more thing: remove the line you added and change this line

```
export PYTHONPATH=$SPARK_HOME/python/lib/py4j-0.8.2.1-src.zip:$PYTHONPATH
```

to

```
export PYTHONPATH=$SPARK_HOME/python/lib/py4j-0.9-src.zip:$PYTHONPATH
```

boazmohar:
@freeman-lab Yep, working as well!

freeman-lab:
Terrific! Let's do it this way, only because Spark technically pins the version of py4j it depends on by bundling it, which is kinda awkward, but we should follow that pattern.

PR #317, opened 2016-05-24
Missing split argument for BoltArraySpark in base.element_wise

### LINKED ISSUE #316: Negative Value Errors for Images
When using the `images.minus()` function, sometimes the values of some pixels may become negative.  

To correct for this, I would like to shift the whole image by a scalar value (The minimum of the difference between the images).  However, after doing the minus call, anytime I try to access the new image object, I get this error:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "build/bdist.linux-x86_64/egg/thunder/images/images.py", line 191, in map
  File "build/bdist.linux-x86_64/egg/thunder/base.py", line 460, in _map
  File "build/bdist.linux-x86_64/egg/bolt/spark/array.py", line 141, in map
  File "build/bdist.linux-x86_64/egg/bolt/spark/array.py", line 94, in _align
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'

Thus, I'm not able to calculate the minimum value across the images to then adjust.

A current workaround is to convert the `image` object to an RDD, calculate the minimum, and adjust the minimum value as an rdd, then do `td.images.fromrdd()` to get back to an RDD.

boazmohar:
fixes #316

### DIFF
--- a/thunder/base.py
+++ b/thunder/base.py
@@ -537,7 +537,7 @@ class Data(Base):
                 return k1, op(x, y)
 
             rdd = self.tordd().zip(other.tordd()).map(func)
-            barray = BoltArraySpark(rdd, shape=self.shape, dtype=self.dtype)
+            barray = BoltArraySpark(rdd, shape=self.shape, dtype=self.dtype, split=self.values.split)
             return self._constructor(barray).__finalize__(self)
 
     def plus(self, other):


boazmohar:
@freeman-lab Something weird happened with the python 3.5 test. Never seen that error before.

freeman-lab:
Yeah me neither, I just restarted the build.

freeman-lab:
Seems to be running normally now...

freeman-lab:
yay!

PR #320, opened 2016-05-25
base.repartition() is missing a sortByKey() to the RDD

### LINKED ISSUE #319: Images object not sorted after repartition
After `repartition()`, calling `first()` gives the wrong record. Probably the RDD need to be sorted.
```python
data = td.images.fromarray(np.ones(shape=( 100, 2, 2)),engine=sc, npartitions=10)
def func(kv):
    k, v = kv
    v = k*v
    return v
data = data.map(func, with_keys=True)
data.first()
out: array([[ 1.,  1.],
       [ 1.,  1.]])
data = data.repartition(25)
data.first()
out: array([[ 20.,  20.],
       [ 20.,  20.]])
```

boazmohar:
fixes #319

### DIFF
--- a/thunder/base.py
+++ b/thunder/base.py
@@ -175,7 +175,7 @@ class Base(object):
             Number of partitions after repartitions.
         """
         if self.mode == 'spark':
-            return self._constructor(self.values.repartition(npartitions), mode=self.mode).__finalize__(self)
+            return self._constructor(self.values.repartition(npartitions)).__finalize__(self)
         else:
             notsupported(self.mode)
 
--- a/thunder/base.py
+++ b/thunder/base.py
@@ -175,8 +175,7 @@ class Base(object):
             Number of partitions after repartitions.
         """
         if self.mode == 'spark':
-            self.values._rdd = self.values._rdd.repartition(npartitions).sortByKey()
-            return self
+            return self._constructor(self.values.repartition(npartitions), mode=self.mode).__finalize__(self)
         else:
             notsupported(self.mode)
 
--- a/thunder/base.py
+++ b/thunder/base.py
@@ -175,7 +175,7 @@ class Base(object):
             Number of partitions after repartitions.
         """
         if self.mode == 'spark':
-            self.values._rdd = self.values._rdd.repartition(npartitions)
+            self.values._rdd = self.values._rdd.repartition(npartitions).sortByKey()
             return self
         else:
             notsupported(self.mode)
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ numpy
 scipy
 scikit-image
 boto
-bolt-python >= 0.6.0
+bolt-python >= 0.7.0
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -113,3 +113,11 @@ def test_repartition(eng):
         assert allclose(data.first(), array([1, 1]))
         data = data.repartition(3)
         assert allclose(data.first(), array([1, 1]))
+
+        data = series.fromlist([array([1, 1]), array([2, 2]), array([3, 3]), array([4, 4]),
+                                array([5, 5]), array([6, 6]), array([7, 7]), array([8, 8]),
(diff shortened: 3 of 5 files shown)


freeman-lab:
Cool! I think we can make this a lot more performant by solving it inside `bolt` though.

There, we use an `ordered` attribute to keep track of whether we know something may have lost its ordering. `sortByKey()` is expensive, and doing it this way means we'll only do it once, during the call to `toarray()`, instead of forcing it after every operation that loses ordering, see https://github.com/bolt-project/bolt/blob/master/bolt/spark/array.py#L993-L1001

There will be a couple steps involved, but should be easy, can be one PR to bolt and then a very simple update to this PR:
- add a check for `ordered` inside the `bolt` spark array's `first()` method just like the one in `toarray()`
- define `repartition` inside `bolt` (seems reasonable, as we already defined `cache` and `unpersist`)
- update this PR to use the `repartition` and `first` from `bolt` inside `thunder`

I'd keep the tests you added here, great to have those!

Sound sensible?

jwittenbach:
@freeman-lab @boazmohar Just to chime in with my two cents: 

Cent #1:
I like the idea of setting `BoltArray._ordered=False` on a `repartition`

Cent #2:
I'm not sure I like the idea of doing it on a `first`. To me, the idea of `RDD.first` is "give me the first example record that you can get your hands on". Is it really worth it to slow that down by demanding that we sort the whole data set first? I'm inclined to say 'no'. The `sortByKey` that happens on a `collect` is essential (thus we definitely need to update `repartition` to set `_ordered=False`), but I'm not convinced that it's the right way to go with `first`.

Thoughts?

boazmohar:
@freeman-lab @jwittenbach I agree that moving it to `bolt` makes sense, I could try to do it today.

Regarding `first`, I have used it both in the sense of "get my a record fast" and " what is the first record".

So maybe a flag for `first` that would force a sort if not sorted would be good.
What are your thoughts on a name and a default behavior as we are venturing far away from the behavior of spark's `first`.

jwittenbach:
I like the idea of a kwarg. Maybe a function signature like:
```python
Data.first(sorted=False)
```
That way the default is to copy the behavior of `RDD.first`, but users still get the option of sorting.

freeman-lab:
Hm, why dooesn't @boazmohar move `repartition` over to `bolt` and we can continue discussing in a PR over there. 

Briefly, I sort of feel that the semantics of `first` on a `bolt` array should be to return... the first item, where first is defined wrt to the fundamental ordering of the array. Also, practically, doing `first()` after a `repartition()` or a `groupBy()` or any of the other operations that disturb sorting carries a big cost anyway -- it's not like doing `first()` after only a `map()` which is always extremely fast -- and the sorting doesn't add that much on top.

boazmohar:
@freeman-lab It passes locally on my machine with the updated bolt.

thunder/base.py
```diff
@@ -175,8 +175,7 @@ def repartition(self, npartitions):
             Number of partitions after repartitions.
         """
         if self.mode == 'spark':
-            self.values._rdd = self.values._rdd.repartition(npartitions).sortByKey()
-            return self
+            return self._constructor(self.values.repartition(npartitions), mode=self.mode).__finalize__(self)
```
boazmohar: @freeman-lab  Is that the right syntax for returning a Base class object?

freeman-lab:
@boazmohar the tests are failing because it's pulling in an earlier `bolt` version, I just published the latest `bolt` to pypi as `0.7.0` if you update `requirements.txt` to specify `bolt >= 0.7.0` that should work and the tests should pass.

thunder/base.py
freeman-lab: very close you shouldn't need the `mode=self.mode` part, it should work just as

```python
return self._constructor(self.values.repartition(npartitions)).__finalize__(self)
```

boazmohar:
@freeman-lab yep, all good. Thanks for the comments!

freeman-lab:
Awesome @boazmohar , looks good to me!

PR #322, opened 2016-05-25
allows series.fromrdd to infer npartitions from shape

### LINKED ISSUE #321: series.frombinary forces count even when conf.json is provided

jwittenbach:
Closes #321


freeman-lab:
LGTM

PR #326, opened 2016-06-02
first returns a BoltArray instead of a ndarray

boazmohar:
added `.toarray()` in images and series, added assertions in the test in base


jwittenbach:
LGTM! Thanks @boazmohar

Issue #327, opened 2016-06-03
SSL error


ObadaM:
I am getting an error after running this line

data = td.images.fromexample('mouse',None)

```
Traceback (most recent call last):
File "/home/l-01-314/PycharmProjects/test/Motion_Correction.py", line 12, in 
data = td.images.frombinary(path=path, npartitions=1, order='F', engine=None)
File "/usr/anaconda3/lib/python3.5/site-packages/thunder/images/readers.py", line 255, in frombinary
buf = reader.read(path, filename=conf).decode('utf-8')
File "/usr/anaconda3/lib/python3.5/site-packages/thunder/readers.py", line 498, in read
storageScheme, key = self.getkey(path, filename=filename)
File "/usr/anaconda3/lib/python3.5/site-packages/thunder/readers.py", line 465, in getkey
scheme, keys = self.getkeys(path, filename=filename)
File "/usr/anaconda3/lib/python3.5/site-packages/thunder/readers.py", line 438, in getkeys
bucket = conn.get_bucket(bucket_name)
File "/usr/anaconda3/lib/python3.5/site-packages/boto/s3/connection.py", line 503, in get_bucket
return self.head_bucket(bucket_name, headers=headers)
File "/usr/anaconda3/lib/python3.5/site-packages/boto/s3/connection.py", line 522, in head_bucket
response = self.make_request('HEAD', bucket_name, headers=headers)
File "/usr/anaconda3/lib/python3.5/site-packages/boto/s3/connection.py", line 665, in make_request
retry_handler=retry_handler
File "/usr/anaconda3/lib/python3.5/site-packages/boto/connection.py", line 1071, in make_request
retry_handler=retry_handler)
File "/usr/anaconda3/lib/python3.5/site-packages/boto/connection.py", line 1030, in mexe
raise ex
File "/usr/anaconda3/lib/python3.5/site-packages/boto/connection.py", line 943, in mexe
request.body, request.headers)
File "/usr/anaconda3/lib/python3.5/http/client.py", line 1083, in request
self._send_request(method, url, body, headers)
File "/usr/anaconda3/lib/python3.5/http/client.py", line 1128, in _send_request
self.endheaders(body)
File "/usr/anaconda3/lib/python3.5/http/client.py", line 1079, in endheaders
self._send_output(message_body)
File "/usr/anaconda3/lib/python3.5/http/client.py", line 911, in _send_output
self.send(msg)
[...]

PR #329, opened 2016-06-07
BotoParallelReader bug 

boazmohar:
Trying to run
`data = td.images.fromexample(name='mouse',engine=None)`
fails due to a change from  #313  as the accessor function expects 3 params.
This PR fixes that, still don't know if related to #327
There are no tests for the BotoParallelReader, seems very hard to mock AWS.
@freeman-lab  Is there another way?

### DIFF
--- a/thunder/readers.py
+++ b/thunder/readers.py
@@ -391,7 +391,7 @@ class BotoParallelReader(BotoClient):
                     idx, keyName = kv
                     key = bucket.get_key(keyName)
                     buf = key.get_contents_as_string()
-                    yield idx, buf
+                    yield idx, buf, keyName
 
             npartitions = min(npartitions, self.nfiles) if npartitions else self.nfiles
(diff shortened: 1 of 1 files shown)


thunder/readers.py
```diff
@@ -391,7 +391,7 @@ def getsplit(kvIter):
                     idx, keyName = kv
                     key = bucket.get_key(keyName)
                     buf = key.get_contents_as_string()
-                    yield idx, buf
+                    yield idx, buf, keyName
```
freeman-lab: make lowercase `keyname`

freeman-lab:
Looks good @boazmohar ! Just a style nit. Thanks for catching this. I think #327 was a different error though.

boazmohar:
Sure thing @freeman-lab I prefer finding the bugs I made first :)

PR #336, opened 2016-06-16
Pushes all map functionality up to Base

jwittenbach:


jwittenbach:
@freeman-lab This was something that I had mentioned doing a while back. It came up presently because I wanted the `value_shape` argument in `map` to avoid the shape/dtype check in `thunder-regression`. However, it wasn't actually exposed on `Series` and was exposed by a different name (`dims`) on `Images`. So I figured I might as well just do it.

The only thing that is slightly awkward is that you can now no longer set the index for `Series` during the `map`, since that doesn't extend to `Images`. I handled this by replacing all instances with the pattern:
```python
return series.map(f, index=index)
```
with
```python
result = series.map(f)
result.index = index
return result
```
which doesn't seem so bad.

jwittenbach:
After offline conversation with @freeman-lab, decided to solve this in a slightly different fashion: #337

PR #337, opened 2016-06-16
updates map API for consistency across Images and Series

jwittenbach:


jwittenbach:
On a `Series` object, if an `index` is not provided, but the mapped function does not change the length of each series, then the old `index` is propagated to the new `Series`.

jwittenbach:
Only API change: the kwarg `dims` on `Images.map` is now `value_shape` for consistency.

jwittenbach:
Closes #338

Issue #335, opened 2016-06-16
Reshape method to compliment flatten for Series objects


bainzo:
Would it be possible to add a `reshape` style method to the Series object?

Currently if you need to flatten a series, for instance to perform factorization, there is no native way of reshaping the resulting series' spatial dimensions to that of the original series.

A workaround involves calling `np.reshape` with a concatenated tuple for the dimensions, for example,
```Python
w_reshape = np.reshape(w, ts.shape[:2] + w.shape[1])
```
However, if the series `w` is on Spark this appears to collect `w` before reshaping, and also returns an Numpy array which requires an extra step to make `w` a series again.

PR #343, opened 2016-06-27
adds reshaping functions to Images and Series

jwittenbach:
implements #335 for `Series` and a corresponding function for `Images`

### DIFF
--- a/thunder/series/series.py
+++ b/thunder/series/series.py
@@ -82,16 +82,8 @@ class Series(Data):
         """
         Reshape all dimensions but the last into a single dimension
         """
-        size = prod([s for i, s in enumerate(self.shape) if i in self.baseaxes])
-        newvalues = self.values.reshape(size, self.shape[-1])
-
-        if self.labels is not None:
-            fullshape = prod(self.labels.shape)
-            newlabels = self.labels.reshape(size, fullshape/size).squeeze()
-        else:
-            newlabels = None
-
-        return self._constructor(newvalues, labels=newlabels).__finalize__(self, noprop=('labels',))
+        size = prod(self.shape[:-1])
+        return self.reshape(size, self.shape[-1])
 
     def count(self):
         """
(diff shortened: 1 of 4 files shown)


jwittenbach:
@freeman-lab: I want to use this to make the handling of non-2D arrays nice and straight-forward in Thunder Factorization. Seemed general enough to include directly in Thunder. Thoughts?

freeman-lab:
Cool, seems reasonable, is this the only case where the thunder API directly references `keys` and `values`? That's the only thing that concerns me a bit. If we've avoided that so far, might not want to introduce it just for this. Other options, off the top of my head:

- You could call both `reshape` but it's confusing if they do different things
- You could only add a `reshape` method to `series`, because for images you can always do reshaping via a `map` operation, but that doesn't work for "reshaping" the keys in `series` data

Other ideas?

jwittenbach:
Yeah, I also thought about just calling it `reshape`, but decided it was too confusing. We do mention "values" in a couple of places (e.g. `value_shape`), but there is no equivalent for "keys", which makes things tricky.

thunder/series/series.py
```diff
@@ -256,6 +256,21 @@ def min(self):
...
+        Parameters
+        ----------
+        shape: one or more ints
+            Shape of the new keys
+        """
+        if prod(self.baseshape) != prod(shape):
+            raise ValueError("Total number of series elements must remain unchanged")
+
+        newshape = shape + (self.shape[-1], )
+        return self._constructor(self.values.reshape(newshape)).__finalize__(self)
```
freeman-lab: should be keys right?

thunder/series/series.py
```diff
@@ -256,6 +256,21 @@ def min(self):
         """
         return self._constructor(self.values.min(axis=self.baseaxes, keepdims=True))
 
+    def reshape_keys(self, *shape):
+        """
+        Reshape the keys of the time series
```
freeman-lab: shouldn't explicitly mention time

thunder/series/series.py
freeman-lab: oh sorry, nevermind, this is correct, i was confused

jwittenbach:
@freeman-lab updated things to reflect our conversation yesterday...both `Images` and `Series` now have a `reshape` function with a simpler API that just takes the new shape, and then catches invalid reshapes.

freeman-lab:
Nice @jwittenbach , this looks great! Merging in.

PR #344, opened 2016-06-28
Fix misspelled "labels" in series.py

WJLiddy:
Tried to convert a series to spark mode and got an error. Looks like labels was just misspelled.


freeman-lab:
ah good catch!

PR #345, opened 2016-06-30
 Add filename to error message when multi page tif has wrong number of planes

boazmohar:
When `discard_extra` is `False` we can now add the filename (as of #313) that it happened in for better debugging.


freeman-lab:
nice addition @boazmohar , LGTM!

PR #346, opened 2016-07-25
Doc fixes in series.convolve

boazmohar:


freeman-lab:
nice, thanks @boazmohar !

PR #348, opened 2016-07-25
bug fix in images.toseries() 'auto' if images length > 100000

### LINKED ISSUE #347: images.toseries() 'auto' setting fails if length is larger then 100,000
```python
data = thunder.images.fromrandom(shape=(110000,10,10,10), engine=sc))
data.toseries()
```
leads to weird chunking as:
```python
if chunk_size is 'auto':
    chunk_size = str(int(100000.0/self.shape[0])))

chunk_size
out: '0'
```

boazmohar:
fixes #347

### DIFF
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -105,7 +105,7 @@ class Images(Data):
         from thunder.series.series import Series
 
         if chunk_size is 'auto':
-            chunk_size = str(int(100000.0/self.shape[0]))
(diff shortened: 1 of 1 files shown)


freeman-lab:
ah, good catch! LGTM, seem ok @jwittenbach ?

jwittenbach:
Yep, makes sense to me -- thanks @boazmohar!

PR #351, opened 2016-08-05
replace skimage with tiffile

freeman-lab:
This PR switches to use `tifffile` for tif writing which means we can remove the dependency on `skimage` (we were only using it for its tif handling).

This means that `images.totif` now supports 8-bit and 16-bit data and both single and multi page tifs.

cc @sofroniewn @boazmohar @d-v-b

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy
 scipy
-scikit-image
-boto
+boto >= 2.36.0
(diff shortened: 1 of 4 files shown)


freeman-lab:
Note that this was only possible because @jwittenbach landed a PR in `tifffile` to enable writing directly to byte streams, otherwise we would have done this awhile ago.

PR #354, opened 2016-08-17
Added a repartition step after reading from tiff in case the number o…

boazmohar:
…f partitions requested is larger then the number of multipage tiff files


boazmohar:
@freeman-lab Hi, do you mind reviewing this? me and @justinplittle were having this problem.
Thanks!

freeman-lab:
@boazmohar @justinplittle great patch! And nice job on the new tests. LGTM, merging in.

PR #355, opened 2016-08-18
'auto' for block_size on series.toimages

### LINKED ISSUE #353: series.toimages()
It seems that there is now a difference between `images.toseries()` and `series.toimages()` default chunk size behavior. Shouldn't they be the same? @freeman-lab @jwittenbach

jwittenbach:
Closes #353

### DIFF
--- a/thunder/images/images.py
+++ b/thunder/images/images.py
@@ -105,7 +105,7 @@ class Images(Data):
         from thunder.series.series import Series
 
         if chunk_size is 'auto':
-            chunk_size = str(max([int(100000.0/self.shape[0]), 1]))
+            chunk_size = str(max([int(1e5/self.shape[0]), 1]))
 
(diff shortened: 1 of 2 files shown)


jwittenbach:
Verified that this fixes the problem that @boazmohar mentioned in #353 -- now calling `Series.toimages` with the default `chunk_size` produces a reasonable number of partitions in the resulting `Images` object. In particular, if the number of pixels is equal to the number of time points ("square matrix"), then `Images.toseries` and `Series.toimages` produce objects with the same number of partitions.

freeman-lab:
Nice fix @jwittenbach , and good job working through the issue @boazmohar ! Seems like a great solution.

jwittenbach:
Thanks @freeman-lab! Merging this in.

PR #356, opened 2016-08-19
Bug fix for images.toblocks()

boazmohar:
When first dimension of the images object is > 1e5 the chunk size is 0 same as #347


jwittenbach:
Absolutely, nice catch @boazmohar!

boazmohar:
@jwittenbach No problem, thank you for #355 now `images.map_as_series` is super fast :)

jwittenbach:
@boazmohar glad to hear that it's running faster!

Issue #341, opened 2016-06-21
load tifs with either tif or tiff extension


freeman-lab:
The default extension for tif files is `tif`, it'd be nice if when loading we automatically looked for either `tif` or `tiff`, as both are quite common in the wild. Currently you need to explicitly set the extension to `tiff`.

PR #357, opened 2016-08-20
Adds tiff files when using images.fromtif default 'tif extension

boazmohar:
Solves #341 
@freeman-lab What is a good default behavior as this is implemented now, it will not enable you to select either tif or tiff from your folders.
I could change the default `ext='tif'` to `'ext='tifs' to make it possible, What do you think?
Also didn't test with s3, any thoughts as to how?

### DIFF
--- a/test/test_readers.py
+++ b/test/test_readers.py
@@ -74,3 +74,24 @@ def test_local_recursive_nested(tmpdir):
     make(tmpdir, filenames)
     actual = LocalFileReader().list(str(tmpdir), recursive=True)
     assert parse(actual) == expected
+
+
+def test_tif_tiff_flat(tmpdir):
+    filenames = ['b.tif', 'a.tif', 'c.tiff']
+    expected = ['a.tif', 'b.tif', 'c.tiff']
+    make(tmpdir, filenames)
+    actual = LocalParallelReader().list(str(tmpdir), ext='tif', recursive=False)
+    assert parse(actual) == expected
+    actual = LocalParallelReader().list(str(tmpdir), ext='tif', recursive=True)
+    assert parse(actual) == expected
+
+
+def test_tif_tiff_recursive(tmpdir):
+    filenames = ['foo/b.tif', 'foo/bar/q.tiff', 'bar/a', 'c.tif', 'd.tiff']
+    expected = ['c.tif', 'd.tiff']
+    make(tmpdir, filenames)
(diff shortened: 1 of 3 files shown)


freeman-lab:
Cool @boazmohar ! I agree this would be good to support, another idea might be to just say if either `tif` or `tiff` is specified, we look for all of them? In other words, make it symmetric, whereas what you have only handles one of the cases. What do you think about that?

test/test_readers.py
```diff
@@ -74,3 +74,24 @@ def test_local_recursive_nested(tmpdir):
     make(tmpdir, filenames)
     actual = LocalFileReader().list(str(tmpdir), recursive=True)
     assert parse(actual) == expected
+
```
freeman-lab: These should probably be in `test_images_io.py`, the `test_readers.py` are for testing generic io functionality relevant to all data types

freeman-lab:
For testing on `s3`, you could at least make sure `thunder.images.fromexample('fish')` works for you after this change, that loads `tif` files from `s3` with the default settings, so is at least a good sanity check for this fix. BTW we don't include that one in the test suits because it takes extra time and requires a network connection.

boazmohar:
@freeman-lab Sure I can make it symmetric but what about a case where someone wants only tif or only tiff files. If you are OK without these options I am too. I thought to have a default of 'tifs' and that would be 'symmetric' as it will include tif and tiff. Specifying another extension will work as expected.
Tell me what you prefer, I don't have a string opinion either way.

test/test_readers.py
boazmohar: I tried moving them but I would need to either import the `make` and `parse` functions or copy paste them into `test_images_io.py`. I think they belong in the `test_readers.py` that is the main function that I changed.

boazmohar:
@freeman-lab fish example is working. Thanks.

test/test_readers.py
freeman-lab: Ah I see, that's fine then.

freeman-lab:
Ah, totally see where you were coming from, I think my call would be to just make it symmetric, as both extensions really do fundamentally specify the same format (see e.g. http://superuser.com/questions/287776/differences-between-tiff-tif-tiff-tif), and it seems very unusual for people to have both in the same location.

boazmohar:
@freeman-lab Thanks for the feedback, ready to merge in my opinion.

freeman-lab:
ok great @boazmohar , merging this in!

Issue #359, opened 2016-09-05
Testing with Spark 2.0


laserson:
I couldn't tell from the README whether Thunder has been tested with Spark 2.0. A Anyone give it a shot?

jwittenbach:
I think @boazmohar has tried this.

boazmohar:
@laserson Yes, I am working with a nightly of Spark 2.0 and thunder is working great with it. I will be tasting the release version this week and will update here with what I find.

freeman-lab:
@boazmohar awesome thanks for confirming, if all the tests run for you locally on Spark 2.0, maybe open a PR that installs 2.0 on travis?

boazmohar:
Sure, I will try later today.

freeman-lab:
Closed by https://github.com/thunder-project/thunder/pull/360

PR #360, opened 2016-09-06
Update for Travis CI to test against spark 2.0

boazmohar:
Address #359

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,12 @@ install:
   - source activate test
   - pip install .
   - pip install -r requirements-dev.txt
-  - wget http://d3kbcqa49mib13.cloudfront.net/spark-1.6.0-bin-hadoop2.6.tgz
-  - tar -xzf spark-1.6.0-bin-hadoop2.6.tgz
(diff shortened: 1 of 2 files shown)


jwittenbach:
Well look at that -- awesome! 👍

freeman-lab:
Well that was straightforward 🎉  thanks @boazmohar ! Can you add a note to the README as part of this PR saying that it works with Spark 2.0.0?

boazmohar:
Yep, was very easy. I guess the parts that changed a lot in 2.0 are `DataFrame` and `Dataset` not `RDD`'s.

lingerlyn:
I've been thinking that it would be interesting to see if we could get any speed-ups by switching to `DataFrames` -- haven't had to play around with it yet though.

jwittenbach:
I've been thinking that it would be test whether there might be any speed-up in switching things to `DataFrames` -- though I haven't had any time to play with it.

freeman-lab:
Yeah agreed @jwittenbach, we should definitely look into that, I think the `Dataset` API may be what we want to look at. But in the meantime let's merge this in, good to make clear it's working with the latest version.

PR #361, opened 2016-09-07
bug fix in base iscached

boazmohar:
fixed a typo
added tests for caching and uncaching


freeman-lab:
Nice fix @boazmohar !

PR #367, opened 2016-12-22
removes pillow from requirements

jwittenbach:
pillow was previously used to write TIFFs, but has since been replaced with tifffile.


jwittenbach:
turns out we actually need pillow because it is required by `scipy.io` but not automatically installed when SciPy is installed -- closing

